PR #41, opened 2017-10-03
Add CsvGenerator.Feature.ESCAPE_QUOTE_CHAR_WITH_ESCAPE_CHAR

frankgrimes97:
Not all CSV parsers accept doubled double-quotes `""` as an escape sequence in a quoted String value.
Added a new feature which allows using the configured escape character instead of hard-coding `""`.
It is false by default to preserve existing behaviour.

### DIFF
--- a/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvGenerator.java
+++ b/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvGenerator.java
@@ -38,7 +38,7 @@ public class CsvGenerator extends GeneratorBase
          *<p>
          * Default value is <code>false</code> for "loose" (approximate, conservative)
          * checking.
-         *
+         * 
          * @since 2.4
          */
         STRICT_CHECK_FOR_QUOTING(false),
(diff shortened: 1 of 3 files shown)


cowtowncoder:
Ah. Yes, I see, and agree in that such an option makes sense.

Since this is an API change, this should ideally go in the next minor version. However...
given that we are about to jump to new major version next, I might make an exception here and allow merging in a patch, since this should be fully backwards compatible. And since otherwise it will probably take quite a while this feature would be available.

I hope to review this in more detail soon, but just want to mention that I have bit of a backlog.

Once again, thank you for contributing this!

cowtowncoder:
Just a ping -- unfortunately I haven't had time to look into this, but will try to get it done for 2.9.3.

cowtowncoder:
@frankgrimes97 Quick question: have I asked for a CLA yet? If not, I'd need one from:

and common way is to print, fill & sign, scan and email to `info` at fasterxml dot com.
Once this is done, I think I can go ahead and merge this as-is.

frankgrimes97:
@cowtowncoder I have sent a signed contributor agreement to the above-mentioned email address.

frankgrimes97:
@cowtowncoder Any update on this?

cowtowncoder:
@frankgrimes97 Sorry, been swamped. I'll try to get this done soon for 2.9.3.

PR #46, opened 2017-10-23
FIX #45 Option to skip ending line break (csv)

### LINKED ISSUE #45: Allow skipping ending line break (`CsvGenerator.Feature.WRITE_LINEFEED_AFTER_LAST_ROW`)
In [RFC-4180](http://www.ietf.org/rfc/rfc4180.txt) it is said that CSV implementations may choose to include a line break at the end of the file or not.

It could be nice to add an option to choose. For instance something like this :

```
schema.withEndingLineSeparator();
schema.withoutEndingLineSeparator();
```

mathieu-lavigne:
Fixes #45.

### DIFF
--- a/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvSchema.java
+++ b/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvSchema.java
@@ -417,6 +417,9 @@ public class CsvSchema
         protected int _escapeChar = DEFAULT_ESCAPE_CHAR;
         
         protected char[] _lineSeparator = DEFAULT_LINEFEED;
+        
+        /** CSV ends with a line separator ? Default is <code>true</code> */
+        protected boolean _endingLineSeparator = true;
 
         /**
          * @since 2.5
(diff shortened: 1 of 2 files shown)


cowtowncoder:
Ok. Thank you for submitting this! It'd be great if logic would prevent outputting instead of trying to undo addition as undo seems bit error-prone. I'll see if that is doable.

mathieu-lavigne:
You're totally right @cowtowncoder. I tried that way first but it seemed to me to be more confusing. One reason is that you have to know when you reach the last line or not.

OmarHawk:
wow, this is open for a very long time.... I actually do need this...

cowtowncoder:
It'd be good to resolve the conflict, could consider merging.

cowtowncoder:
@mathieu-lavigne Apologies for dropping the ball on this. I think this would work, as `CsvEncoder.endRow()` does guarantee that linefeed is not split on output buffer boundary. There may be some edge cases still (someone calling `CsvGenerator.flush()` at the end before `close()`...) but that is probably acceptable considering how difficult alternative is (that of trying to change how linefeed is added).

But I'd like one bigger change: instead of `CsvSchema`, I think this really works better as `CsvGenerator.Feature`.
There's also need for a unit test or two.

With those I'd be happy to merge this for 2.17.

I realize you may not have time or interest after this delay; if not, I understand. And can then proceed with a new PR.

cowtowncoder:
Created new PR #453 to implement as generator feature.

Found couple of issues wrt implementation -- `generator.flush()` will prevent it from working, basically -- but aside from that works ok (added couple of basic tests).

Issue #68, opened 2018-02-06
When field names are reserved words, they should be written out with quotes


baconmania:
https://github.com/FasterXML/jackson-dataformat-yaml/blob/master/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLGenerator.java specifies that YAML keys will always be written with `STYLE_NAME`, e.g. without quotes.
```java
    // for field names, leave out quotes
    private final static Character STYLE_NAME = null;
...
    private final void _writeFieldName(String name)
        throws IOException
    {
        _writeScalar(name, "string", STYLE_NAME);
    }
```

But this breaks parsers when you write a map with a key that is a reserved word in YAML, e.g. `no`. If `_writeFieldName()` encounters a field that is one of `{ y, true, yes, on, n, false, no, off }`, it should use `STYLE_QUOTED`.

Actual definition of such boolean values can be found here:

https://yaml.org/type/bool.html

cowtowncoder:
Thank you for reporting this, contributing patch.

Bit odd that SnakeYAML does not do escaping. But as with everything, looks like YAML has gone off the deep end with the rather long list of reserved names.... so maybe it's impractical.

baconmania:
Thanks for the quick response!

And yeah, YAML's actual spec is [chaos](http://www.yaml.org/spec/1.2/spec.html):

```txt
The content of a mapping node is an unordered set of key: value node pairs, with the restriction that
each of the keys is unique. YAML places no further restrictions on the nodes. In particular, keys may be
arbitrary nodes, the same node may be used as the value of several key: value pairs, and a mapping
could even contain itself as a key or a value (directly or indirectly).
```

Given that, the minimal-change way to support reserved-words keys without actually implementing the full YAML spec in Jackson's writer is adding a feature to just quote all keys.

But hey, I'll ask you one favor: would you able to backport this to jackson-dataformat-yaml 2.7.x (or at least 2.7.9)?

cowtowncoder:
@baconmania Wrt 2.7, unfortunately no. If I followed semantic versioning strictly, this could even go in 2.9 being new feature. But since that branch is open and likely the last 2.x version I don't have problem doing that. However backporting is different story -- 2.7 in particular is closed and only "micro-patches" are to be release (like 2.7.9.1) for critical bugs.

Thank you for feature PR itself; I'll add just one note there.

cowtowncoder:
[...]

PR #69, opened 2018-02-06
Add feature to support quoted field names in YAML writer.

baconmania:
Fixes https://github.com/FasterXML/jackson-dataformats-text/issues/68.

/cc @cowtowncoder


cowtowncoder:
Looks good, and although adding features in patches is usually no-no, I think I'll make exception here because 2.9 will probably be the last 2.x version (and conversely 3.0 will take quite a while to be released).

Just one process thing: unless I've asked for and received it, we need CLA before the first merged contribution. It's one time, one page document from:

and usual way is to print, fill & sign, scan, email to `info` at fasterxml dot com.
Once it is received I can merge this PR as well as any other contributions.
Apologies for the hassle.

Thank you again for reporting this & contribution the fix!

cowtowncoder:
@baconmania hiya! I would really like to merge this, but would need CLA if you get a chance.

cowtowncoder:
Implemented in slightly different, as per #68.

PR #78, opened 2018-04-17
Fix Jackson reference

WilderPereira:


cowtowncoder:
Not quite sure what the fix here would be... relative reference resolves to the same as proposed change?

WilderPereira:
No, it doesn't. It is currently resolving to https://github.com/FasterXML/jackson-dataformats-text/jackson

cowtowncoder:
Ah. My mistake: was looking at main-level README, not csv one.

cowtowncoder:
Thanks!

PR #79, opened 2018-04-20
Update README.md

dcbartlett:
When attempting to implement mapper.readValues, the parameter was vague and i passed in the file name directly.  This could be solved by explaining what csvFile variable is.


cowtowncoder:
Yes. Thank you! Documentation fixes, updates are very much welcome, READMEs are too sparse as is so I appreciate help.

Issue #84, opened 2018-05-01
Add option to allow use of platform-linefeed (`YAMLGenerator.Feature.USE_PLATFORM_LINE_BREAKS`)


thauk-copperleaf:
The behaviour I expect would be for the serializer to produce output with line endings consistent with that platform it's running on, unless explicitly overriden.

I'm just wondering if this is on purpose or not?

Deeper dive:

`YAMLGenerator` uses `DumperOptions` from snakeyaml and overrides some settings, but does not override the line ending setting:

https://github.com/FasterXML/jackson-dataformats-text/blob/a0f5c80bb431dbba9a9f505c3f1554aa4be0bfc5/yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLGenerator.java#L231-L255

And since `DumperOptions` has the following default:
```    private LineBreak lineBreak = LineBreak.UNIX;```

That results in output having Unix line endings, regardless of platform.

Alternatively, can you tell me how to produce YAML with platform-specific line endings?

cowtowncoder:
Generator uses SnakeYAML defaults so this is not conscious definition. I actually do not know what YAML specs says: I assume it probably allows multiple styles of linefeeds just like JSON and XML.
Whether platform defaults should be used or canonical defaults is an open question; since both are (I presume) legit I don't have strong opinion.

I would be open to having an option to change behavior: being an API change, it needs to go in Jackson 3.0 however.

One challenge there is this: it is easy to add on/off kinds of things (`YAMLGenerator.Feature`).
However, with 3.0 it will be much easier to add backend-specific configuration via Builder-style construction. But to keep dependencies to `SnakeYAML` hidden, we'd probably need a new enumeration for choices as well.
That seems cleaner than adding logic to detect platform, although maybe that'd be fine as an alternative. But that would then require definition of some kind of universal default, for feature for "enable platform-specific linefeed" (if disabled, what is the linefeed then -> default).

thauk-copperleaf:
@cowtowncoder 

Thanks for the quick reply!

> I actually do not know what YAML specs says: I assume it probably allows multiple styles of linefeeds just like JSON and XML.

Correct. The [YAML spec](http://yaml.org/spec/1.1/#id871136) allows four different "generic line breaks", including CRLF for Windows and LF for Unix.

> Whether platform defaults should be used or canonical defaults is an open question; since both are (I presume) legit I don't have strong opinion.

[...]

PR #86, opened 2018-05-08
Add USE_PLATFORM_LINE_BREAKS option to YAMLGenerator.Feature

thauk-copperleaf:


thauk-copperleaf:
Addresses https://github.com/FasterXML/jackson-dataformats-text/issues/84

cowtowncoder:
That works. Thank you for submitting the PR!

One wish: would be great to have a test, if(f) it can be implemented reliably (which I would assume is possible by getting linefeed snakeyaml uses, or maybe just `System.getProperty("line.separator")`?

Aside from that, would be happy to merge. All I need is Contributor License Agreement (CLA), from:

and usually easiest way is to print, fill  & sign, scan, email to `info` at fasterxml dot com, and we are good to go.

PR #93, opened 2018-06-23
Avoid escaping when writing to a properties object

### LINKED ISSUE #91: `JavaPropsGenerator#writeFieldName()` should not escape property keys
I think that ```JPropEscapes.appendKey(_basePath, name);``` in ```JavaPropsGenerator#writeFieldName``` should be moved *only* to ```WriterBackedGenerator```. Escaping keys is a serialization matter when a Properties object is written to something other than memory. So, if you do:

```
public class TestObject {
	Map<String, String> values = new HashMap<>();
	public Map<String, String> getValues() {
		return values;
	}
	public void setValues(Map<String, String> values) {
		this.values = values;
	}
}
[...]
@Test
	public void test() throws Exception {
		TestObject expected = new TestObject();
		expected.values.put("foo:bar", "1");
		Properties properties = mapper.writeValueAsProperties(expected);
		TestObject actual = mapper.readPropertiesAs(properties, TestObject.class);
		assertThat(actual.values).isEqualTo(expected.values);
	}
```

This fails because the properties object already contains the key escaped as ```foo\:bar```, but readProperties expects that it doesn't. The properties itself, will add another backslash if it's about to be written to an actual file.

dmandalidis:
Fix for #91


cowtowncoder:
Manually merged.

PR #97, opened 2018-07-29
Add serialization of object members for csv.

ptahchiev:
- Keep track of object nesting, similar to jackson-dataformat-properties

### DIFF
--- a/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvEscapes.java
+++ b/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvEscapes.java
@@ -1,0 +1,99 @@
+package com.fasterxml.jackson.dataformat.csv;
+
+import java.util.Arrays;
+
+/**
+ * @author Petar Tahchiev
+ * @since 2.0.1
+ */
+public class CsvEscapes {
+    private final static char[] HEX = "0123456789ABCDEF".toCharArray();
+
+    private final static int UNICODE_ESCAPE = -1;
+
+    private final static int[] sValueEscapes;
+    static {
+        final int[] table = new int[256];
+        // For values, fewer escapes needed, but most control chars need them
+        for (int i = 0; i < 32; ++i) {
+            table[i] = UNICODE_ESCAPE;
+            // also high-bit ones
+            table[128+i] = UNICODE_ESCAPE;
+        }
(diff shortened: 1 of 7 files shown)


nstdio:
Is there any problem with this PR?

cowtowncoder:
@nstdio At this point there is one practical problem (aside from merge conflict): being new feature, this needs to be rebased against 2.11. But I do not remember any specific problem against the concept: my main question would be that since `@JsonUnwrapped` allows similar usage, whether overlap is problematic. Maybe it isn't, but functionality-wise this covers similar case -- however, I can see how with multi-format output, not using that annotation would be preferable.

I will add this PR to my todo list to have another look as it seems valuable, in helping deal with nested values in CSV.

nstdio:
@cowtowncoder yes, exactly. Also, users not always control/own models and adding a `@JsonUnwrappd` annotation might be not an option.

cowtowncoder:
Ok. So, finally getting to this one.... couple of problems actually.

A practical one is that unfortunately patch has re-styled code, making diff unnecessarily large, and adding work when I have to undo many (not all, just many) changes.
Since I will need to rebase it to be from 2.11, manually, these are bit unfortunate.

But a bigger one is this: I don't want additional processing overhead for common case of no nesting. So will need to figure out how to avoid automatic prepending and only change name in cases where it is needed for nesting.

So this may take a bit longer yet.

cowtowncoder:
Another thing I am not sure about is addition of path separator per column; not sure how that would work compared to per document (one for whole `CsvSchema`) would work. Reformatting is especially disruptive for `CsvSchema` in general.

branaway:
I'd love to see this feature goes in the code. since CVS is much more compact than JSON in dealing with large data set. In my case I have to convert between json and CSV without any predefined POJO involved. This is much needed and I don't want to create a fork out of the official Jackson

cowtowncoder:
Note: would like to consider this, but would need rework to merge against 2.13 (next applicable minor version).

cowtowncoder:
Ok so due to unfortunate reformatting of this patch (probably accidentally done by some reformatter?), it's bit difficult to follow, but now realized that this would only support serialization, not deserialization.
Not quite sure if that makes sense on its own; users might be surprised to find they can serialize things but not deserialize.

Issue #100, opened 2018-08-04
Add an option to specify properties prefix


alonbl:
Hi,

Properties may have a prefix for an object, for example:
```
a.b.c.name =
a.b.c.id =
x.y.z.name =
x.y.z.color =
```

It would be nice if a prefix can be specified, so that only these properties with the prefix will be loaded, prefix will be stripped and then business as usual. When writing, the prefix will be added.

This will allow reading system properties and complex file properties using this interface quite easily and replace many of the complex properties processing.

If this is acceptable, I can workout a patch.

Thanks!

dharezlak:
Can this be configured somehow with the current configuration options? I am interested in writing properties with a common prefix to a file.

alonbl:
I could not find, a workaround:

```java
import com.fasterxml.jackson.annotation.*;
import com.fasterxml.jackson.databind.*;
import com.fasterxml.jackson.dataformat.yaml.*;
import java.util.*;

public class T {

     static class Test {
           @JsonProperty
           int field1 = 44;
     }

     public static Object nestObject(Object o, List<String> levels) {
           if (levels.size() == 0) {
                return o;
           }
           else {
                return Collections.singletonMap(levels.get(0), nestObject(o, levels.subList(1, levels.size())));
           }
     }

     public static Object nestObject(Object o, String... levels) {
           return nestObject(o, Arrays.asList(levels));
     }

     public static void main(String... args) throws Exception {
           ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
           System.out.println(mapper.writeValueAsString(nestObject(new Test(), "a", "b", "c")));
     }
}
```

dharezlak:
I successfully used `withRootName()` for the writing case:

```
JavaPropsMapper mapper = new JavaPropsMapper();
mapper
	.writerFor(Bean.class)
	.with(JavaPropsSchema
			.emptySchema()
			.withKeyValueSeparator(" = ")
			.withWriteIndexUsingMarkers(true)
			.withFirstArrayOffset(0)
			.withIndexMarker(Markers.create("[", "]")))
	.withRootName("root.subroot")
	.writeValueAsString(bean);
```

cowtowncoder:
This sounds reasonable to me; useful thing. I would be interested in a PR -- ideally against `2.10` branch, since this seems like an incremental feature that need not wait until 3.0.

I could see this working either via reuse of "root name" concept, or, as a new property of `JavaPropsSchema`.

[...]

PR #105, opened 2018-08-31
properties: add properties prefix as filter/append

alonbl:
Properties may contain multiple objects based on prefix, for example:

    a.b.c.name =
    a.b.c.id =
    x.y.z.name =
    x.y.z.color =

To read these formats a prefix should be used as a filter when reading and as
append when generating.

A new optional JavaPropsSchema::prefix() and JavaPropsSchema::withPrefix() are
added to control the behavior.

As each path component is an object, use of ObjectReader::withRootName has a
different meaning so schema property was used.

PR #100


cowtowncoder:
CLA received, will need to figure out how to merge in 2.10.

PR #111, opened 2018-10-20
fix #90 do not escape line separator in yaml long binary fields

### LINKED ISSUE #90: Exception when decoding Jackson-encoded `Base64` binary value in YAML
Using jackson-dataformats-text-2.9.6 binary values that exceed a given length cannot be decoded back and the following exception is thrown:

```
com.fasterxml.jackson.core.JsonParseException: Illegal character '\' (code 0x5c) in base64 content
```

This is related to #62 which forces the `Base64Variant` to be `MIME`. When the encoded value exceeds a given size, line seperators are inserted into the encoded value but it then fails to be decoded.

I created the following small test and added it to the `SimpleGenerationTest` in `yaml`:

```java
public void testCase() throws Exception {
        final int length = 50;
        final byte[] data = new byte[length];
        Arrays.fill(data, (byte) 1);

        final YAMLFactory yamlFactory = new YAMLFactory();

        final ByteArrayOutputStream os = new ByteArrayOutputStream();
        try (JsonGenerator generator = yamlFactory.createGenerator(os, JsonEncoding.UTF8)) {
            generator.writeStartObject();
            generator.writeBinaryField("data", data);
            generator.writeEndObject();
        }

        try (JsonParser parser = yamlFactory.createParser(os.toByteArray())) {
            assertEquals(JsonToken.START_OBJECT, parser.nextToken());
            assertEquals(JsonToken.FIELD_NAME, parser.nextToken());
            assertEquals("data", parser.currentName());
            assertEquals(JsonToken.VALUE_EMBEDDED_OBJECT, parser.nextToken());
            assertArrayEquals(data, parser.getBinaryValue());
            assertEquals(JsonToken.END_OBJECT, parser.nextToken());
            assertNull(parser.nextToken());
        }
    }
```

This test passed correctly when `length` is small (`50`) but with larger size (like `100`) it fails.

vboulaye:
fixes #90 by copying the encode method from Base64Variant in jackson-core and changing the way it print line separators (they are used to split the base64 output in 76 characters lines):
- line separators are not escaped (the escaping of '\n' made it impossible for the yaml doc to be parsed
- uses the line separator defined according tothe USE_PLATFORM_LINE_BREAKS Feature from #84

### DIFF
--- a/yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLGenerator.java
+++ b/yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLGenerator.java
@@ -19,8 +19,7 @@ import com.fasterxml.jackson.core.base.GeneratorBase;
 import com.fasterxml.jackson.core.json.JsonWriteContext;
 import com.fasterxml.jackson.core.io.IOContext;
 
-public class YAMLGenerator extends GeneratorBase
-{
+public class YAMLGenerator extends GeneratorBase {
     /**
      * Enumeration that defines all togglable features for YAML generators
      */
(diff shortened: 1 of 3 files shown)


cowtowncoder:
@vboulaye thank you for submitting this! I used the idea and just merged manually; fix will be in 2.9.9 and 2.10.0. Especially appreciate the tests here.

Issue #124, opened 2019-03-27
Allow escaping of non-printable characters in CSV output/input


hgschmie:
We have a use case where we replace an existing CSV writer (based on Apache Commons CSV) with Jackson. The old CSV writer was configured to write "special characters" such as CR and LF as `\r` and `\n`. Jackson does not support this but adheres to RFC 4180 (where no escaping exists). This causes a lot of pain for our customers as the data we write contains often CR and LF  characters. 

Here is some test code:

```java
public class CsvTest {

    public static final void main(String... args) throws Exception {

        CsvSchema schema = CsvSchema.emptySchema().withEscapeChar('\\');
        CsvFactory csvFactory = new SchemaAwareCsvFactory(schema);

        ObjectMapper csvMapper = new ObjectMapper(csvFactory);

        String [] line = new String [] { "a", "\n", "\"", ","};

        csvMapper.writeValue(new PrintWriter(System.out), line);

    }

    // Is there actually a better way to set the schema for a ObjectMapper? This seems painful.
    public static class SchemaAwareCsvFactory extends CsvFactory {
        SchemaAwareCsvFactory(CsvSchema schema) {
            super();
            this._schema = schema;
        }
    }
}
```

Which produces

```
a,"
","""",","
```

I can get it to produce 

```
"a","
","\"",","
```
by adding

```java
csvFactory.enable(CsvGenerator.Feature.ALWAYS_QUOTE_STRINGS);
csvFactory.enable(Feature.ESCAPE_QUOTE_CHAR_WITH_ESCAPE_CHAR);
```

But what I am actually looking for is

```
"a","\n","\"",","
```

There seems to be no way to get the generator (and probably also the parser) to generate and parse control characters. Having CR and LF within quotes is legal from the RFC 4180 PoV, however most of the CSV that our systems produce get parsed by legacy ("brain dead") tools that assume that every LF is a record separator.

Apache Commons CSV has a nice summary on their Javadoc page for `CSVFormat`: https://commons.apache.org/proper/commons-csv/apidocs/org/apache/commons/csv/CSVFormat.html#DEFAULT (and below)

(god, CSV is such a mess. And that is the *standard* format for enterprise data???)

cowtowncoder:
Yes, I think this makes sense -- and yes, CSV is a mess, making this tricky.

The main challenge is really that of figuring API extensions that would be needed.
[...]

PR #125, opened 2019-04-01
Allow escaping of non-printable characters in CSV output/input

hgschmie:
This is a proposed solution for #124. It introduces a new Feature, `ESCAPE_CONTROL_CHARS_WITH_ESCAPE_CHAR`,
which will apply the standard ASCII escapes from JSON to all characters that the CSV generator writes.

If this solution is workable, I will add tests.


hgschmie:
The same patch also applies cleanly to the 2.10 code branch.

hgschmie:
@cowtowncoder ping? Any chance for you to weigh in / reject here?

cowtowncoder:
Ok, merged, will be in 2.9.9.

Forgot to ask for a unit test to verify, but that can be added at a later point.

PR #136, opened 2019-06-11
Note SnakeYAML Engine in README

grv87:


cowtowncoder:
Thank  you for contribution. 

One note: new SnakeYAML engine will only be used in 3.0, so not sure what to do about that. 2.x uses "old" snakeyaml (or at least 2.9 does, and 2.10 almost certainly -- if there is 2.11 that's open question).
So changing description at this point would require mention that in future implementation will change.

grv87:
Engine is already in current master, which this PR is against: https://github.com/FasterXML/jackson-dataformats-text/blob/master/yaml/pom.xml#L32

cowtowncoder:
@grv87 yes. And current master is not used for Jackson 2.x, but for next major version, 3.0. Which will not be available for months. So for all existing users (and near term releases), "old" SnakeYAML is used, not the new version.

So all I am saying is that since most users look at default README, but may not be aware of how branches map to versions, it'd be good to indicate that actual engine used varies.

grv87:
As a user I was investigating the issue with boolean values not parsed correctly. When I found out from pom that snakeyaml-engine is used but README said different I was very confused. This is why I made this PR.

I've added new commit, please review.

cowtowncoder:
Perfect, thank you! I'll see if I can add something more about branches; but this should help new users/devs a lot.

grv87:
@cowtowncoder, you're welcome :)
Please note that README should be updated as soon as 3.0 is out.

cowtowncoder:
Yes, duly noted. Many things will change at that point, need to figure out how to indicate differences between 2.x and 3.x.

PR #143, opened 2019-09-09
Fix #83: Update index on context in array.

### LINKED ISSUE #83: (yaml) Update index of sequence context
I've a problem with a custom deserializer which relies on `JsonStreamContext.getCurrentIndex()` to reflect the current collection index. The YAMLParser fails to update the `_index` field and thus the following fails for YAML but completes successfully for JSON:

```
import com.fasterxml.jackson.annotation.JsonProperty;
import com.fasterxml.jackson.core.JsonParser;
import com.fasterxml.jackson.databind.DeserializationContext;
import com.fasterxml.jackson.databind.JsonDeserializer;
import com.fasterxml.jackson.databind.ObjectMapper;
import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
import org.junit.Test;

import java.io.IOException;

import static org.junit.Assert.assertArrayEquals;

public class Sequence2Test {
    private static class IdDeser extends JsonDeserializer<Integer> {
        @Override
        public Integer deserialize(JsonParser p, DeserializationContext ctxt) {
            return p.getParsingContext().getCurrentIndex();
        }
    }

    private static class MySequence {
        @JsonProperty @JsonDeserialize(contentUsing = IdDeser.class)
        Integer[] myList;
    }

    @Test
    public void testSequenceIteratorYaml() throws IOException {
        testSequence(new ObjectMapper(new YAMLFactory()));
    }

    @Test
    public void testSequenceIteratorJson() throws IOException {
        testSequence(new ObjectMapper());
    }

    private void testSequence(ObjectMapper mapper) throws IOException {
        MySequence seq = mapper.readValue("{\"myList\": [9,8,7]}", MySequence.class);
        assertArrayEquals(new Integer[] { 0, 1, 2 }, seq.myList);
    }
}
```
I guess it's a bit awkward to detect start of a new item (which might be scalar or object), but maybe someone more familiar with how this works for JSON can suggest a fix? Or is this working as intended?

mwinkels:
Fix #83

### DIFF
--- a/yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLParser.java
+++ b/yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLParser.java
@@ -400,6 +400,11 @@ public class YAMLParser extends ParserBase
                     return (_currToken = JsonToken.FIELD_NAME);
                 }
             }
+
+            if (_parsingContext.inArray()) {
+                _parsingContext.expectComma();
+            }
+
             // Ugh. Why not expose id, to be able to Switch?
             _currentAnchor = null;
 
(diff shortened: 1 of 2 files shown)


cowtowncoder:
@mwinkels Thank you very much for contributing this!

I also appreciate your fixing this against version branch. However, due to 2.9 being late patch, I want to minimize possibility of breakages further, so I think I will actually merge this just for 2.10 -- I know this should be low risk, but I have been bitten by changes in the past.
2.10.0 should be out relatively soon so that should not delay getting fix out much (in fact, it'll go out before 2.9.10).

cowtowncoder:
As per notes manually merged, fix will be in 2.10.0.

PR #153, opened 2019-10-05
fixes #15 SKIP_EMPTY_LINES feature for csv parser

### LINKED ISSUE #15: (csv) Add a `CsvParser.Feature.SKIP_EMPTY_LINES` to allow skipping empty rows
Looks like a commonly supported option with CSV (family of formats) is to allow skipping of empty rows; that is, ones that only contain white-space.

vboulaye:
I started an implementation of the SKIP_EMPTY_LINES feature based on the existing ALLOW_COMMENTS.
When enabled It ignores both empty and blank lines.

Do you prefer to have SKIP_EMPTY_LINES and a new SKIP_BLANK_LINES as separate features? or is it good enough to adapt the description of SKIP_EMPTY_LINES?


cowtowncoder:
I don't think 2 features are needed, but I think `SKIP_BLANK_LINES` might be better name -- this because there are a few priori cases like `String.isEmpty()` that specifically do not check for "all whitespace" case.

vboulaye:
I renamed the feature to `SKIP_BLANK_LINES` (and all the references in the code). 

Is there something to do to handle the deprecation since the old name `SKIP_EMPTY_LINES` has already been published?

I also made a change to the `CommentsTest` to include a test case where the '#' character is preceded only by whitespaces. This is a  case that is described in `CsvSchema.setAllowComments` but was not explicitly tested.

Regarding the documentation, I did not modify the javadoc html as I assumed it is generated, but I could not find from where.

cowtowncoder:
Oh shoot. I did not remember that `SKIP_EMPTY_LINES` did exist. I thought that was being added.

Now: since support had not been implemented, I'd have been ok to change it for 2.10.0, but since that ship has sailed, I think we have 2 choices:

1. Use `SKIP_EMPTY_LINES`, implement for 2.10.1 (no api change, just fix to make existing feature work
2. Replace `SKIP_EMPTY_LINES` with `SKIP_BLANK_LINES` for 2.11(.0), as API change, deprecating old feature.

To me (1) seems like a better choice actually; we do need to document it, but that's fine.
Then we could do renaming for 3.0.

Apologies for not checking this before asking you to change this.

csv/src/main/java/com/fasterxml/jackson/dataformat/csv/impl/CsvDecoder.java
```diff
@@ -65,12 +69,14 @@
     protected boolean _trimSpaces;
 
     protected boolean _allowComments;
-    
+
```
cowtowncoder: Need to add `@since`, once we figure which branch this should go in. Same for new methods.

vboulaye: I added `@since 2.10` to the feature  + the new method that replace the skip comments.
I am not sure where else I should put it.

vboulaye:
I reverted the name of the Feature to SKIP_EMPTY_LINES and added  ̀@since 2.10` as it was added in 2.10 anyway

csv/src/main/java/com/fasterxml/jackson/dataformat/csv/impl/CsvDecoder.java
cowtowncoder: I usually add to method and field, but not a big deal for internal fields.

cowtowncoder:
Ugh. There's a pretty nasty bug in here. Code fails to take into account the fact that only part of the document may be in buffer, leading to #191 ... :-( :-( :-(.

PR #155, opened 2019-10-09
Use index of Mark object instead of -1

### LINKED ISSUE #151: `YAMLParser._locationFor()` does not use index available from `Mark` object of Event
`YAMLParser._locationFor` [uses -1 for the `totalChars` parameter](https://github.com/FasterXML/jackson-dataformats-text/blob/770ef8403c2e672789fb5dcef66fd066f6432ebb/yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLParser.java#L270) when creating a new `JsonLocation`. However, it has an `index` property which can be used just perfectly for the `totalChars` parameter.

yfklon:
- Fixes #151

### DIFF
--- a/yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLParser.java
+++ b/yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLParser.java
@@ -267,7 +267,7 @@ public class YAMLParser extends ParserBase
         }
         Mark m = option.get();
         return new JsonLocation(_ioContext.getSourceReference(),
-                -1,
+                m.getIndex(),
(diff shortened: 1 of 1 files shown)


cowtowncoder:
Fix looks reasonable, thank you for contributing it!

Would it be possible to add a unit test to verify that index value is sensible?

Aside from that, I'm happy to merge this PR, but need just one more thing: CLA (unless I have received one already -- once filed, it's good for any contributions for all Jackson repos). It can be found here:

and the easiest way is to print, fill & sign, scan/take photo, email to `info` at fasterxml dot com.
Once I receive that, I can merge this PR.

Thank you again!

PR #159, opened 2019-11-10
CsvMapper now honors JsonProperty index.

### LINKED ISSUE #115: JsonProperty index is not honored by CsvSchema builder
In the case if I don't want to use JsonPropertyOrder, which looks to be less convenient than use "index" property:
Properties are ordered alphabetically, (regardless of their "index" values), so the test failing with 2.9.7
According to the index documentation, it should be used for schema generation:
"Property that indicates numerical index of this property (relative
     * to other properties specified for the Object). This index
     * is typically used by binary formats, but may also be useful
     * **for schema languages and other tools**."

```
public class CsvTest {
    public static class Pojo {
        @JsonProperty(required = true, value = "Name", index = 1)
        private String name;
        @JsonProperty(required = true, value = "Abba", index = 2)
        private int ab;
    }

    @Test
    public void tt() throws IOException {
        String csvText = "Name,Abba\n" + "Alex, 1\n";

        CsvMapper mapper = new CsvMapper();
        CsvSchema schema = mapper
                .schemaFor(Pojo.class)
                .withHeader()
                .withStrictHeaders(true);
        List<Pojo> out = mapper.readerFor(Pojo.class).with(schema).<Pojo>readValues(csvText).readAll();
        assertEquals(out.size(), 1);
    }
}
```

nstdio:
Resolves #115

### DIFF
--- a/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvMapper.java
+++ b/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvMapper.java
@@ -1,6 +1,11 @@
 package com.fasterxml.jackson.dataformat.csv;
 
+import static java.util.Comparator.comparing;
+import static java.util.Comparator.naturalOrder;
+import static java.util.Comparator.nullsLast;
+
 import java.util.Collection;
+import java.util.List;
 
 import com.fasterxml.jackson.core.type.TypeReference;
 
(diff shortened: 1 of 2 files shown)


cowtowncoder:
Thank you for submitting this! I will try to get this reviewed soon.

csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvMapper.java
```diff
@@ -459,7 +464,13 @@ protected void _addSchemaProperties(SerializerProvider ctxt, CsvSchema.Builder b
         }
         BeanDescription beanDesc = ctxt.introspectBeanDescription(pojoType);
         final AnnotationIntrospector intr = ctxt.getAnnotationIntrospector();
-        for (BeanPropertyDefinition prop : beanDesc.findProperties()) {
+        List<BeanPropertyDefinition> properties = beanDesc.findProperties();
+
+        // soring properties by @JsonProperty index
+        properties.sort(comparing(o -> intr.findPropertyIndex(ctxt.getConfig(), o.getPrimaryMember()),
```
cowtowncoder: Jackson 2.x requires JDK 7, can not yet use JDK 8 features.

cowtowncoder:
Finally got back to this. Will use provided test, but also decided that this:

https://github.com/FasterXML/jackson-databind/issues/2555

makes sense, to essentially try to use `index` in general (to avoid/reduce need for same code everywhere). Since there is some limited risk for breakage will do for 2.11 (and not in 2.10.2 patch). After getting databind done, will see if more work needed on CSV side or not.

cowtowncoder:
Turns out databind change does solve the problem, so I only had to merge test from here to verify that things now work. Thank you for contributing the test and proposed fix!

PR #183, opened 2020-03-17
Fixes #182. Negative numbers serialized with quotes.

### LINKED ISSUE #182: Negative numbers not serialized correctly
When initializing a `YamlFactory` as follows, negative numbers are not quoted correctly.

```
YAMLFactory factory = new YAMLFactory();
factory.enable(YAMLGenerator.Feature.MINIMIZE_QUOTES);
factory.enable(YAMLGenerator.Feature.ALWAYS_QUOTE_NUMBERS_AS_STRINGS);
```

Serializing `60` currently results in `"60"`.
Serializing `-14` currently results in `14`.

dpbevin:
Fixes #182


bartveenstra:
When can we expect 2.10.4?

cowtowncoder:
@bartveenstra It will take a while as 2.10.3 was released 2 weeks ago. At this point in patches there are usually 1 - 3 months between. Not too many fixes accumulated yet, either:

https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.10.4

so perhaps in late April. Problem with releases is that every full minor release set takes hours to do (less than it used to, but still 2-3 hours if all goes well).

PR #192, opened 2020-04-08
Fix #191

### LINKED ISSUE #191: `ArrayIndexOutOfBoundsException` when skipping empty lines, comments
Hi we found a bug in the csv parsing.

For some csv files parsing fails with:

`java.lang.ArrayIndexOutOfBoundsException: Index 4000 out of bounds for length 4000
	at com.fasterxml.jackson.dataformat.csv.impl.CsvDecoder.skipLinesWhenNeeded(CsvDecoder.java:527)
	at com.fasterxml.jackson.dataformat.csv.impl.CsvDecoder.startNewLine(CsvDecoder.java:499)`

I did some digging it happens if there is a newline as char 4000 in the csv file.

Happens in versions > 2.10.0

I think it was introduced with #15

Can i help with a fix?

f-julian:
added failing testcase


cowtowncoder:
@f-julian I'd be happy to actually just merge this, see if I can fix the underlying problem? I typically add failing tests under `failing` package (which will be skipped during regular build, only manually run from IDE), and once fix is in, move to somewhere else.

cowtowncoder:
Merging, assuming this test fails before fix.

PR #195, opened 2020-04-23
Adds schema creating csv schema with View

Qnzvna:


cowtowncoder:
@Qnzvna this looks very useful, thank you! I think I have some comments wrt implementation, but in general I think I like the idea.

Just one practical thing before I can merge this: unless I have asked for CLA:

I would one before first contributions (but only then; future contributions can all be under existing CLA then).
Usually the easiest way is to print, fill & sign, scan/photo, email to `info` at fasterxml dot com.

csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvMapper.java
```diff
@@ -460,6 +487,18 @@ protected void _addSchemaProperties(SerializerProvider ctxt, CsvSchema.Builder b
         BeanDescription beanDesc = ctxt.introspectBeanDescription(pojoType);
         final AnnotationIntrospector intr = ctxt.getAnnotationIntrospector();
         for (BeanPropertyDefinition prop : beanDesc.findProperties()) {
+            if (view != null) {
+                boolean viewVisible = false;
+                for(Class<?> propView : prop.findViews()) {
```
cowtowncoder, line 492: Needs null check, `findViews()` can return `null`. Also note that in databind `BeanDeserializerFactory` method `addBeanProps()`, there is:

```
                Class<?>[] views = propDef.findViews();
                if (views == null) {
                    views = beanDesc.findDefaultViews();
                }
```

csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvMapper.java
```diff
@@ -460,6 +487,18 @@ protected void _addSchemaProperties(SerializerProvider ctxt, CsvSchema.Builder b
...
+                    if (propView.equals(view)) {
```
cowtowncoder, line 493: Not sufficient: Views actually check inheritance too (so, say, `CharSequence` as inclusion would be acceptable for View `String`).
You might be able to use databind's `ViewMatcher` for checks:

```
public static ViewMatcher construct(Class<?>[] views) { }
```

and then calling method:

```
        public boolean isVisibleForView(Class<?> activeView) { ...}
```

seems like the proper (and still simple) way to go.

csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvMapper.java
Qnzvna: Thanks, updated.

csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvMapper.java
Qnzvna: Thanks, updated also.

Qnzvna:
I've signed the CLA before when providing different contributions.

cowtowncoder:
@Qnzvna Ahh! I did find the CLA from 2017. Thank you!

cowtowncoder:
Excellent: I backport this in 2.11 branch and this will just make it in 2.11.0 thanks to timing.

Thank you again for contributing this new feature!

Qnzvna:
Thanks for merging so quickly.

cowtowncoder:
@Qnzvna and 2.11.0 is actually out now! Perfect timing. :)

PR #196, opened 2020-04-24
Upgrade to SnakeYAML Engine 2.1

asomov:
Upgrade SnakeYAML and fix issue with parsing the leading plus (+1000.25)


asomov:
Sorry, I forgot that it was already applied

PR #181, opened 2020-03-12
FIX #180: YAML Generator now quotes strings containing special chars

trohrberg:
When using the YAMLGenerator in the MINIMIZE_QUOTES mode, the generator serializes strings in plain style, i.e. without quoting them. However, as stated in the YAML Spec [1] such plain style strings cannot contain some special characters such as ':', ',', and others.

[1] https://yaml.org/spec/1.2/spec.html#id2788859

As with the serialization of boolean-like strings, the YAMLGenerator should be adjusted to always quote strings containing those special characters.

I will shortly try to provide a pull request to fix this issue.

Issue #180, opened 2020-03-12
YAMLGenerator serializes string with special chars unquoted when using `MINIMIZE_QUOTES` mode


trohrberg:
When using the YAMLGenerator in the MINIMIZE_QUOTES mode, the generator serializes strings in plain style, i.e. without quoting them. However, as stated in the [YAML Spec](https://yaml.org/spec/1.2/spec.html#id2788859) such plain style strings cannot contain some special characters such as ':', ',', and others.

As with the serialization of boolean-like strings, the YAMLGenerator should be adjusted to always quote strings containing those special characters.

I will shortly try to provide a pull request to fix this issue.

PR #201, opened 2020-06-03
REF #180 - updates YAML scalars with indicators quoting style

frantuma:
#181 and 2 more #180 related commits introduce some improved YAML scalars quoting more compliant to the [YAML Spec](https://yaml.org/spec/1.2/spec.html#id2788859), in the MINIMIZE_QUOTES mode; the updated behavior seems to be at least more strict (in quoting) than necessary, as it introduces quoting for any value containing chars as `#` and `:`.

If I am not mistaken [YAML Spec](https://yaml.org/spec/1.2/spec.html#id2788859) states that such plain style limitation is applied for cases where the character combinations ` #` (`<space_or_tab>+<#>`) and `: ` (`<:>+<space_or_tab>`) are present, not the single chars `#` and `:`.

This PR updates behavior by considering the character combinations and not the single chars.

This has been triggered by an update of Jackson from 2.10.4 to 2.11.0 which caused YAML serialization test failures due to the difference in quoting, e.g. in case of scalar representing URIs, JSON references, date or datetime.

I understand that this is probably not a major issue, and e.g. code or tests can be adjusted in dependent projects and downstream, but I believe the affected scenarios are quite common and it would make sense to keep quotes at a minimum when `minimize` is on.

Not sure if the target branch `2.11` is the correct one, was wondering if this could make it for 2.11.1 patch release in case.

### DIFF
--- a/yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLGenerator.java
+++ b/yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLGenerator.java
@@ -175,7 +175,7 @@ public class YAMLGenerator extends GeneratorBase
      * aliases for booleans, and we better quote such values as keys; although Jackson
      * itself has no problems dealing with them, some other tools do have.
      */
-    // 02-Apr-2019, tatu: Some names will look funny if escaped: let's leave out 
+    // 02-Apr-2019, tatu: Some names will look funny if escaped: let's leave out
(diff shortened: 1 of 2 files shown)


cowtowncoder:
@frantuma Thank you! I think you nailed it; I like the idea of more targeted quoting, and I think risk/reward ratio works for 2.11.1 (early enough patch). Will merge.

PR #225, opened 2020-09-16
quote special boolean characters when writing

pnepywoda:
"y" is a boolean (https://yaml.org/type/bool.html) and should be quoted
when written, just like all the other special boolean characters

### DIFF
--- a/yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLGenerator.java
+++ b/yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLGenerator.java
@@ -1,25 +1,21 @@
 package com.fasterxml.jackson.dataformat.yaml;
 
-import java.io.*;
-import java.math.BigDecimal;
-import java.math.BigInteger;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.Map;
-import java.util.Set;
-import java.util.regex.Pattern;
-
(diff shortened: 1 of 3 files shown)


cowtowncoder:
Hmmmh. As per comment

                // NOTE: single-letter cases left out on purpose

this seems like a conscious decision, not an omission. But based on that if change was to be made, it would need to go in a minor version (so 2.12 branch instead of 2.11). This because changes like these tend to result in bugs getting reported (and sometimes breakages by users who rely on specific behavior (brittle code and they probably shouldn't)).

I should probably reach out `jackson-dev` and/or `jackson-users` google group to see how others feel about this.
One possibility would be to add a feature for `YAMLGenerator.Feature`, but then again this is sort of of a minor detail that maybe that'd be an overkill.

So, let me think about this a bit.

pnepywoda:
The conscious decision's justification was "Some names will look funny if escaped"

I agree there's a chance this breaks something, so thanks for opening the discussion on if we push back on that

cowtowncoder:
Right that is what I wrote -- but I am trying to remember if there was more to the story than that.
I suppose this is one issue where Release Candidates (2.12.0-rc1) can help; to expose changes when they can still be reverted or improved (wrt configurability). I am almost thinking that it'd make sense to consider pluggable handler to let users configure such quoting aspects.

But for now what I will do is I think I will apply changes to 2.12; will create separate issue and merge manually.

cowtowncoder:
Created #226 to manually merge the fix. Thanks!

Issue #185, opened 2020-03-26
Unable To Serialize Yaml To Object Format With Jackson


bhreinb:
Hi there,

This is either a question or minor issue (apologies if it's the former). I'm serializing a pojo using the below code logic

```
public static <T> String SerializeToRawYaml(final T yamlT) {        
       final ObjectMapper mapper = new ObjectMapper(new YAMLFactory()
                .configure(YAMLGenerator.Feature.WRITE_DOC_START_MARKER, false)
        )
                .enable(SerializationFeature.WRITE_SINGLE_ELEM_ARRAYS_UNWRAPPED);
        mapper.setVisibility(mapper.getSerializationConfig().getDefaultVisibilityChecker()
                .withFieldVisibility(JsonAutoDetect.Visibility.ANY)
                .withGetterVisibility(JsonAutoDetect.Visibility.NONE)
                .withSetterVisibility(JsonAutoDetect.Visibility.NONE)
                .withCreatorVisibility(JsonAutoDetect.Visibility.NONE)
                .withIsGetterVisibility(JsonAutoDetect.Visibility.NONE));
        try {
            return mapper.writerWithDefaultPrettyPrinter().writeValueAsString(yamlT);
        }
        catch (JsonProcessingException e) {
            System.out.println(
                    String.format("Problem Serializing POJO Because Of Error %s", e)
            );
        }
    }
```

This produces sample yaml like so 

```
name: "OuterYamlElementName"
label: "OuterYamlElementLabel"
fields:
  label: "outerYamlFieldLabel"
  value: "outerYamlFieldValue"
  isRequired: true
  toggle: "+BEYWAYS-Outer-Yaml-Field"
elements:
  name: "InnerYamlElementDepthOneName"
  label: "InnerYamlElementDepthOneLabel"
  fields:
  - label: "innerYamlFieldsDepthOneItemOneLabel"
    value: "innerYamlFieldsDepthOneItemOneValue"
    isRequired: true
    toggle: "-BEYWAYS-Inner-Yaml-Fields-Depth-One-Item-One"
  - label: "innerYamlFieldsDepthOneItemTwoLabel"
    value: "innerYamlFieldsDepthOneItemTwoValue"
    isRequired: false
  toggle: "-BEYWAYS-Inner-Yaml-Element-Depth-One"
toggle: "+BEYWAYS-Outer-Yaml-Element"
```

however I'd like the yaml to be generated like so

```
name: "OuterYamlElementName"
[...]

Issue #133, opened 2019-05-20
Document distinction of YAML multiple documents vs sequence/array of values


cowtowncoder:
(see #132 for background)

So: handling of a single physical document containing multiple logical documents is quite specific in that one has to use `mapper.readValues()` to deal with it, and can not use regular `readValue()` to bind as collections (for example). There are technical reasons for this difference but it is not necessarily clear from user perspective this is a different use case.

So this should be somehow documented. An immediate place is probably on README for YAML module, but perhaps something could be added to `readValues()` method too, to mention it does not just related to root-level JSON values (like "linefeed-separate JSON") but also more generally to some other data formats.

Issue #43, opened 2017-10-17
Option to use square brackets for arrays instead of dash


sscarduzio:
Hi, is there a way to get this:
```yaml
    kibana_hide_apps: ["readonlyrest_kbn", "timelion"]
```

instead of this:
```yaml
    kibana_hide_apps:
    - "readonlyrest_kbn"
    - "timelion"
```

cowtowncoder:
I assume this is for YAML backend. At this point I don't think there is no `YAMLFeature` for this, although I think SnakeYAML (which is underlying decoder/encoder) should be able to do that -- it is difference between "inline" vs "block" notation I think.

It would probably make sense to allow selecting this just for arrays (and perhaps another one for Objects), if adding a new feature.

Another possibility would be to allow somehow passing specifically configured SnakeYAML emitter/parser, as this would expose full range of settings.

cowtowncoder:
If anyone knows configuration within `SnakeYAML`, I'd be interested in figuring out how to pass this configuration.

asomov:
Is [this](https://bitbucket.org/asomov/snakeyaml/src/cea5dd3eb8e124e547ad62d9919e4b9a891c4992/src/test/java/examples/CollectionStyleTest.java?at=default&fileviewer=file-view-default) what you are looking for ?
FlowStyle.BLOCK, FlowStyle.FLOW, FlowStyle.AUTO

AUTO means that the first will be block style and then flow style.

cowtowncoder:
Thank you @asomov. That helps understanding intent between styles.

Can this be changed during generation, or does the initial setting remain in effect when generation starts?

asomov:
it remains

cowtowncoder:
No plans to implement, closing.

PR #242, opened 2021-02-03
Add FlowStyleResolver to enable custom YAML node style resolution

axpendix:
This adds a desired mechanism to specify YAML node styles (flow styles in snakeyaml) for individual object and array nodes, via implementing a FlowStyleResolver and checking the field names of given object or array nodes.

References: #185 #133 #43

### DIFF
--- a/yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLFactoryBuilder.java
+++ b/yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLFactoryBuilder.java
@@ -1,5 +1,6 @@
 package com.fasterxml.jackson.dataformat.yaml;
 
+import com.fasterxml.jackson.dataformat.yaml.util.NodeStyleResolver;
 import org.yaml.snakeyaml.DumperOptions;
 
 import com.fasterxml.jackson.core.TSFBuilder;
@@ -34,6 +35,13 @@ public class YAMLFactoryBuilder extends TSFBuilder<YAMLFactory, YAMLFactoryBuild
     protected StringQuotingChecker _quotingChecker;
 
     /**
+     * Helper object used to determine node styles of objects and arrays.
+     *
+     * @since 2.13
+     */
+    protected NodeStyleResolver _nodeStyleResolver;
+
+    /**
      * YAML version for underlying generator to follow, if specified;
      * left as {@code null} for backwards compatibility (which means
      * whatever default settings {@code SnakeYAML} deems best).
@@ -55,6 +63,7 @@ public class YAMLFactoryBuilder extends TSFBuilder<YAMLFactory, YAMLFactoryBuild
         _formatGeneratorFeatures = base._yamlGeneratorFeatures;
         _version = base._version;
         _quotingChecker = base._quotingChecker;
+        _nodeStyleResolver = base._nodeStyleResolver;
     }
 
     // // // Parser features NOT YET defined
@@ -117,6 +126,12 @@ public class YAMLFactoryBuilder extends TSFBuilder<YAMLFactory, YAMLFactoryBuild
         return this;
     }
 
+
+    public YAMLFactoryBuilder nodeStyleResolver(NodeStyleResolver nodeStyleResolver) {
+        _nodeStyleResolver = nodeStyleResolver;
+        return this;
+    }
+
     /**
      * Method for specifying YAML version for generator to use (to produce
      * compliant output); if {@code null} passed, will let {@code SnakeYAML}
@@ -150,6 +165,10 @@ public class YAMLFactoryBuilder extends TSFBuilder<YAMLFactory, YAMLFactoryBuild
             return _quotingChecker;
         }
         return StringQuotingChecker.Default.instance();
+    }
+
+    public NodeStyleResolver nodeStyleResolver() {
+        return _nodeStyleResolver;
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLFactoryBuilder.java
```diff
@@ -117,6 +126,12 @@ public YAMLFactoryBuilder stringQuotingChecker(StringQuotingChecker sqc) {
         return this;
     }
 
+
+    public YAMLFactoryBuilder nodeStyleResolver(NodeStyleResolver nodeStyleResolver) {
```
cowtowncoder, line 130: Could you add `@since 2.13` javadoc to this (and getter) too?

axpendix: done :)

cowtowncoder:
@axpendix Apologies for slow follow up here: will try to get this done soon now (hopefully today). Quick question: did I mention CLA earlier? If not, I'd need this:

and after I get it, can merge. CLA only needs to be sent once so if this was already taken care of no need to re-send (I just need to locate it).
If you do need to send it, the common way is to print, fill & sign, scan/photo, email to `info` at fasterxml dot com.

Looking forward to merging this!

yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/util/NodeStyleResolver.java
```diff
@@ -0,0 +1,73 @@
...
+    };
+
+    /**
+     * Resolve a node style for given fieldName.
+     *
+     * @param fieldName parent field name of the current object or array. can be null if there is no parent field (i.e.
+     *                  typically root object)
+     * @return the desired {@link NodeStyle} or null to use default value (currently 'BLOCK')
+     */
+    NodeStyle resolveStyle(String fieldName);
```
cowtowncoder, line 71: I realized that instead of passing just `fieldName`, it'd be more flexible to pass write context; and perhaps also have 2 methods to call: one for Objects, another for Arrays.
I guess it would be possible to pass both `name` (current name in context) and context, to simplify some cases, while giving full power for others.

At very least I think array and object cases should go to different methods since I think that's relatively common use case.

yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLGenerator.java
```diff
@@ -517,13 +524,16 @@ public final void writeStartArray() throws IOException
...
         _verifyValueWrite("start an array");
         _writeContext = _writeContext.createChildArrayContext();
-        FlowStyle style = _outputOptions.getDefaultFlowStyle();
         String yamlTag = _typeId;
         boolean implicit = (yamlTag == null);
         String anchor = _objectId;
         if (anchor != null) {
             _objectId = null;
         }
+        NodeStyle jacksonStyle = _nodeStyleResolver.resolveStyle(
```
cowtowncoder, line 533: Would make sense to use `getCurrentName()` bit earlier here, before `createChildArrayContext()`.

Would this work with nested arrays? (might be that it gets `null`).

Also: would make sense to get and pass default style to custom resolver -- but resolver could be passed the context as well (pre-addition of array), to let it access all information, including ancestors. I'll add a note on interface later on.

yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/util/NodeStyleResolver.java
cowtowncoder: ... just realized I added this note already. Wrt comment I just added above.

And looking at this method now, maybe passing the default style is not useful after all (because it would need to be translated etc). Returning `null` makes sense.

cowtowncoder:
@pjfanning What do you think about this earlier PR? I don't think I have gotten a CLA to merge this, but I suppose that exposing more of SnakeYAML's functionality might make sense. Not sure how well it'd merge with snakeyaml-engine in 3.0 but otherwise.

pjfanning:
@cowtowncoder I haven't really messed about with these features of snakeyaml. The only reason I wanted to expose the LoaderOptions was because the default 3Mb limit for parsing is quite small.

cowtowncoder:
@pjfanning sure, understood. And changes here are bit more involved as the handler can change settings on per-property name basis, which should be tested etc.

pjfanning:
@cowtowncoder one thing we could do is to allow users to set their own DumperOptions on YAMLFactory (like they can now with LoaderOptions). If users provide their own non-null DumperOptions, then https://github.com/FasterXML/jackson-dataformats-text/blob/2.14/yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLGenerator.java#L294 will be skipped. Would this make sense as v2.14 change?

cowtowncoder:
@pjfanning I think that is a good starting point and probably covers many use cases even if not that particular one.

asomov:
@axpendix do you think that something can be improved in SnakeYAML to support it ?
[Is this similar ?](https://bitbucket.org/snakeyaml/snakeyaml-engine/issues/24/allow-flowstyle-property-to-be-configured)

asomov:
no progress also in SnakeYAML:
https://bitbucket.org/snakeyaml/snakeyaml-engine/issues/24/allow-flowstyle-property-to-be-configured

Issue #116, opened 2019-01-24
Error handling "null" String when Feature.MINIMIZE_QUOTES is active


stefanleh:
When serializing a String with content "null" and Feature.MINIMIZE_QUOTES is active the "null" value does not get quoted. Therefore when deserializing the value is not a String but **_null_**.

In com.fasterxml.jackson.dataformat.yaml.YAMLGenerator.writeString(String) there is a special handling for boolean content like "true" and "false". Therefore id suggest to add special handling for "null" and "NULL" string values there as well.

What do you think about that?

cowtowncoder:
Sorry, I had missed this one. Adding on my W-I-P list. Sounds like worthy addition, yes.

cowtowncoder:
Thank you for reporting this, suggesting the fix. I manually merged fix in `2.10`, will be in 2.10.0

PR #246, opened 2021-03-05
Special characters shouldn't force double quoting for multi-line strings

ahgittin:
the logic added in c427465219bbb7e728ac6378b3ab21fd6dccf151 for #116 to force double-quoting of strings should only apply to single-line strings.  the quoting of multi-line strings look naff, and because they take literal mode, the characters aren't special and the quotes are not necessary.

### DIFF
--- a/yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLGenerator.java
+++ b/yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLGenerator.java
@@ -591,13 +591,13 @@ public class YAMLGenerator extends GeneratorBase
         DumperOptions.ScalarStyle style;
         if (Feature.MINIMIZE_QUOTES.enabledIn(_formatFeatures)) {
             // If one of reserved values ("true", "null"), or, number, preserve quoting:
-            if (_quotingChecker.needToQuoteValue(text)
+            if (text.indexOf('\n') >= 0) {
+                style = STYLE_LITERAL;
+            } else if (_quotingChecker.needToQuoteValue(text)
                 || (Feature.ALWAYS_QUOTE_NUMBERS_AS_STRINGS.enabledIn(_formatFeatures)
                         && PLAIN_NUMBER_P.matcher(text).matches())
                 ) {
                 style = STYLE_QUOTED;
-            } else if (text.indexOf('\n') >= 0) {
-                style = STYLE_LITERAL;
             } else {
                 style = STYLE_PLAIN;
             }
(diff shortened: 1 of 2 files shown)


ahgittin:
BTW ideally the characters `[]{},` should only be treated as special if used in a key or in a flow collection [1].  They are permitted in strings in plain style when used as a value in a map.  Not working on that now but thought I'd raise it.

[1] https://yaml.org/spec/1.2/spec.html#id2788859

cowtowncoder:
Apologies for slow follow-up here. Yes, I think you are right.

On bracket/curly-braces case... I suspect one should only have to worry about opening ones, too? Since it's only pairs that matter, and so if there's no `{` or `[`, it should not matter if there's `]` or `}`?
Except that perhaps that, too, depends on Key/Object-value/Array-Value difference.

Now, it is possible to distinguish key handling (it's separate method already); and we know from context Array/Object difference as well (so could/should be moved to separate method too?).

With that I assume that linefeeds should still force quoting of keys for sure.

Anyway. I hope to get back to this one soon and hope we can figure out improvements. Quoting/escaping has proven to be a rather problematic thing....

cowtowncoder:
Patch itself is simple: but it would probable make sense to add this to handling of `StringQuotingChecker`. Not 100% sure how the flow should go. I also realize there is bit of backwards-compatibility challenge, but since this was only introduced in 2.12, I think there's slightly lower risk of breakage if we have to change something.

Specifically I see that no info is passed on context (Array vs Object vs Root value (not sure if we have root values in YAML but theoretically)). Passing `JsonWriteContext` as extra argument would give that information for code that cares; or, alternatively could split `needToQuoteValue()` into 2 methods.

cowtowncoder:
Sigh. Quoting/style checking in 2.12.x is a royal mess.

I think this specific change makes sense, and I think I'll actually backport it in 2.12.3.

But for 2.13 (and esp. 3.0) this really should be rethought. I also realize that check for key quoting is not considering "funny characters" at all, which is probably not correct... although not sure what SnakeYAML at low level will do.

cowtowncoder:
As per above comment, will manually merge suggested fix for 2.12. Also filed #252 to improve `StringQuotingChecker` so that users could conceivably override handling to their liking (currently abstraction is too limited to allow that).

cowtowncoder:
As per above noted, merged manually in 2.12 branch, for inclusion 2.12.3. Thanks!

ahgittin:
I agree with all your points -- esp that keys do still need quoted and that makes life more painful.  Thanks for fixing this!

cowtowncoder:
@ahgittin Thank you for the fix! I will be releasing 2.12.3 patch today.

PR #251, opened 2021-04-07
Toml merge 2.13

yawkat:


cowtowncoder:
Excellent, thank you!

aalmiray:
@cowtowncoder @yawkat once again, thank you for this amazing (and very quick!) work.

Just ran a test against the JReleaser model format and it works without a hitch!

```toml
[project]
  name = "app"
  description = "Awesome App"
  longDescription = "Awesome App"
  website = "https://acme.com/app"
  authors = ["Duke"]
  license = "Apache-2.0"
[project.java]
  groupId = "com.acme"
  version = "8"

[release.github]
  owner = "aalmiray"
  overwrite = true
  sign = true

[signing]
  enabled = true
  armored = true

[distributions.app]
  artifacts = [
    { path = "target/distributions/{{distributionName}}/{{distributionName}}-{{projectVersion}}.zip" }
  ]
```

Turns out anonymous objects (such as the _artifact_ element in _artifacts_) work out of the box as well. Neat!

cowtowncoder:
All credit really goes to @yawkat -- this is pretty awesome success so far.

One thing that occurred to me: while officially this format module will be announced for 2.13, we could quite easily sneak it in upcoming 2.12.3. The reason to do that would be just make early testing easier as it will be some time until first release candidate of 2.13 will be out.
Whereas 2.12.3 will be released much sooner, possibly within a week but definitely during April 2021.

aalmiray:
Having this new artifact as part of 2.12.3 sounds great to me. I've been holding off publishing JReleaser for a few days due to "issues" (GH actions got disabled after the recent spam a few days ago). I can certainly wait for 2.12.3 or just add TOML support a few days/weeks later.

cowtowncoder:
Cool. I think I'll do this, should be just simple cherry-pick.

Somewhat ironically, the last (?) open piece wrt 2.12.3 is a thing with YAML output style escaping :-D

cowtowncoder:
Ok took bit more work since I forgot `ContentReference` work I've done, but now backported to 2.12 as well.
Will do some more warnings removal on 2.13 but really happy to get this all in.

cowtowncoder:
One more question, @yawkat, on naming (yes I am sometimes obsessed with that :) ). So, capitalization/acronym, should we use `TOML` or `Toml` for class names? Code as-is uses "Toml", which is fine; 
I realize I have not been consistent at all within Jackson since while mostly this is the pattern:

* `CsvMapper`, `JsonMapper`, `XmlMapper`

there is an exception (of course):

* `YAMLMapper`

So `TomlMapper` etc makes sense. But if we did want to use all-caps (I think name itself does that, as per https://en.wikipedia.org/wiki/TOML), now would be the last easy time to do it.

So... anyone feeling strongly that we ought to use full-caps "TOML" instead of "Toml"?

yawkat:
I used `Toml` because that's the google camel case style: https://google.github.io/styleguide/javaguide.html#s5.3-camel-case

cowtowncoder:
Ok. Since this is more common case within Jackson codebase anyway, let's go with it. Just thought it's better to briefly discuss now & be done with it.

toml/src/main/java/com/fasterxml/jackson/dataformat/toml/TomlWriteContext.java
```diff
@@ -0,0 +1,198 @@
...
+        _inline = (parent != null && parent._inline) || type == TYPE_ARRAY;
+    }
+
+    private void reset(int type, Object currValue, int basePathLength) {
+        _type = type;
+        _basePathLength = basePathLength;
+        _currentValue = null;
+        _index = -1;
+        _currentValue = currValue;
+    }
```
cowtowncoder, line 72: @yawkat I suspect that `_inline` will need to be recalculated as well? Have not created a test (for a while was considering #258 might be due to this, but that is separate) but thought I'd point this out.

Basically, with reuse, if initially context was Object, reused as Array (or vice versa), this likely would be incorrect.
But before changing would be good to reproduce actual problem. :)

cowtowncoder: Fixed as per #260.

PR #259, opened 2021-04-09
Fix #258 (no output with byte-backed targets)

### LINKED ISSUE #258: `TomlMapper.writeValueAsBytes()` (and similar) fails to produce output
Looks like there is something wrong with byte output of `TomlMapper` / `TomlFactory`: String/character output is fine but for some reason byte buffering seems to not get any content.

I can reproduce the issue but trying to find the fix while I'm at it.

cowtowncoder:
As per title: fix is to pass standard generator features in addition to toml ones; this will correctly pass `flush()` calls as necessary.
(there's potential open question on whether we really should just force flush since we are creating the buffering writer, but for now this suffices).


cowtowncoder:
@yawkat I'll merge this in 2.12, merge forward, I hope it does not mess with your changes too badly.

PR #263, opened 2021-04-10
Use buffer recycler for TOML buffers

yawkat:
This PR makes the jflex-generated Lexer use allocTokenBuffer for the main token buffer (ba4f8edd47b42daf913447b50d20ae90b1c77759) and constructTextBuffer for string building (c042f4e5f6943eaad1fba1145ecc31362c039d83).

This is accomplished by altering the default jflex skeleton ( https://github.com/jflex-de/jflex/blob/4edfabcaa7eb06a41f8429a7462eca6dc92d7247/jflex/src/main/resources/jflex/skeleton.default ) to remove the character array allocations, and implementing those using IOContext in the .jflex instead. Needing to copy the whole skeleton isn't ideal, but oh well.

This approach only works if individual jflex tokens are smaller than the default buffer returned by allocTokenBuffer. When there are very long tokens, we fall back on new char[]. We cannot use allocTokenBuffer(minSize), because we need to keep the old buffer available, and there cannot be two token buffers in use at the same time from the same IOContext.

Ideally, we would never need to resize the buffers at all. Some of the potential long tokens (maybe all?) are tested in LongTokenTest.

- long numbers could maybe just throw an exception. For json, jackson writes long numbers to the text buffer as fallback, but any "fallback" behavior that only happens with long inputs is difficult with jflex, because DFAs can't count. It'd require creating a new state that implements this behavior.
- comments and whitespace could be ignored, since we never need to access the content of those tokens. Could be solved with a new state fairly easily, if jflex cooperates.
- strings are already copied to the text buffer, and this logic could be improved. Right now, long sequences of unescaped data are processed as a single token for performance, but maybe there's an alternative here too. The "fallback" behavior problems for numbers apply.

All in all I'm not awfully concerned about this problem. None of these cases are likely to occur for "benign" data; even string and comment data is regularly "flushed" as long as there's a newline somewhere in between. So, the only issue here is that someone could build a malicious TOML file that takes a lot of memory to parse, but even then the attack is not very effective, because at most 4x as much memory is used as there is data (2x for utf-8 -> char array, 2x for the resize factor). There are definitely better attacks, e.g. by creating many json nodes (since we cache the whole tree during parsing).

For this reason, imo, it is not worth addressing this beyond what is already in this PR.

### DIFF
--- a/toml/LICENSE
+++ b/toml/LICENSE
@@ -1,5 +1,7 @@
 Jackson is licensed under [Apache License 2.0](http://www.apache.org/licenses/LICENSE-2.0.txt).
 
+---
+
 The TOML grammar and some parts of the test suite are adapted from the TOML project, which is licensed under the MIT license:
 
 The MIT License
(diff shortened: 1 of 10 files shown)


yawkat:
jackson-benchmark results for the 2.13 version of this patch:

before:
```
Benchmark                                    Mode  Cnt       Score       Error  Units
TOMLStdReadAfterburner.readPojoMediaItem    thrpt    7  123674.553 ±  2572.701  ops/s
TOMLStdReadVanilla.readPojoMediaItem        thrpt    7  119795.915 ±  1100.263  ops/s
TOMLStdWriteAfterburner.writePojoMediaItem  thrpt    7  496205.741 ±  7262.247  ops/s
TOMLStdWriteVanilla.writePojoMediaItem      thrpt    7  469790.847 ± 11204.921  ops/s
```

after:
```
Benchmark                                    Mode  Cnt       Score      Error  Units
TOMLStdReadAfterburner.readPojoMediaItem    thrpt    7  134938.605 ± 4316.725  ops/s
TOMLStdReadVanilla.readPojoMediaItem        thrpt    7  129000.369 ±  994.874  ops/s
TOMLStdWriteAfterburner.writePojoMediaItem  thrpt    7  515883.030 ± 7075.571  ops/s
TOMLStdWriteVanilla.writePojoMediaItem      thrpt    7  475328.670 ± 2755.866  ops/s
```

Probably spent more time building this patch than would ever be saved by the improvement :D

yawkat:
Oh, one thing that was a bit weird @cowtowncoder , I had to use `textBuffer.emptyAndGetCurrentSegment();` instead of just `resetWithEmpty`, even though I think both are supposed to be equivalent. resetWithEmpty sets the currentSegment to null, leading to an NPE in any subsequent append calls. Am I using it wrong or is that a bug?

cowtowncoder:
@yawkat First of all, thank you for figuring this out! I was wondering if this was possible.

I also recognize the feeling of doing some optimization, only to find its effects are less than what profiler might have suggested. Still, +10% is not too shabby in general and sometimes there's nice stacking effect at some point (some combination of smaller changes manages to compound to something bigger).

I'll have a look at what's up with `TextBuffer`, there may be some odd corner case with initialization, some assumption code has on call patterns. `resetWithEmpty()` should usually be sufficient in cases where value of "empty String" is implied. But `emptyAndGetCurrentSegment()` is the expected call when starting to build new Strings.

cowtowncoder:
Ok, actually, code as it is now is correct. `resetWithEmpty()` javadocs could probably use improvement as it really should only be used to indicate the case of "current text collected is empty String" -- but it is not reset to start where collection can begin. Essentially all `resetXxx()` methods mean "set buffer to contain value X".

But one thing I could consider here is ... optimizing `TextBuffer.append()`. Most existing usage just gets underlying `char[]` segment for direct modification, but that does not work with Lexer hear, as it's state machine based, and `append()` makes more sense. I suspect I could at least try changing things a bit to streamline the happy path and see if I can observe actual improvement.
Usually I might not bother but I suspect this is a hot spot now, on critical path. And code in question is small enough that I can play with it, run tests etc. If it doesn't matter should be easy to find out. :)

yawkat:
Alright, good to know. I'm going to merge this tomorrow, then.

cowtowncoder:
Sounds good, looking forward to merged version!

cowtowncoder:
Oh. Should have looked at profiler output bit more... I don't think single-char `TextBuffer.append()` is actually used a lot.
Most usage via bulk appends. Will see if something can be done there, still. Maybe there's some room for optimization.

Another sort of interesting thing is that on reader side, as all input will become `ObjectNode` at least briefly, it might be possible to make `ObjectMapper.readTree()` avoid tree-from-tree construction.
The place to do this would probably be `JsonNodeDeserializer` which could detect use of `TreeTraversingParser` and then just use the copy? The only question is whether use of custom sub-class (of ObjectNode/ArrayNode) might be problematic or not.
Not 100% sure how much overhead this would avoid, potentially not a lot... but there is some overhead in traversal and recreation of the whole tree.

PR #271, opened 2021-05-19
Fix quote of # character in strict mode, fix #270

### LINKED ISSUE #270: [jackson-dataformat-csv] Should not quote with strict quoting when line starts with #, but comments are disabled
When STRICT_CHECK_FOR_QUOTING is enabled and withoutComments() is used, Jackson should not escape strings starting with "#"

```
CsvMapper mapper = new CsvMapper();
mapper.configure(CsvGenerator.Feature.ALWAYS_QUOTE_STRINGS, false);
mapper.configure(CsvGenerator.Feature.STRICT_CHECK_FOR_QUOTING, true);

CsvSchema schema = mapper.schemaFor(recordClass).withoutComments();
mapper.writerFor(recordClass).with(schema).writeValue(wrtiter, record);
```

The record contains a single string field with value #abc.

Expected:
#abc is written to CSV file

Actual:
"#abc" is written to CSV file

In CsvEncoder::_needsQuotingStrict "(c == '#' && i == 0)" should be checked only when comments are enabled.

kdebski85:
Should not quote with strict quoting when line starts with #, but comments are disabled. Quote only the first column, fix #270

### DIFF
--- a/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/impl/CsvEncoder.java
+++ b/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/impl/CsvEncoder.java
@@ -73,7 +73,9 @@ public class CsvEncoder
      * @since 2.5
      */
     final protected char[] _cfgNullValue;
-    
+
+    final protected boolean _cfgAllowsComments;
+
     final protected int _cfgLineSeparatorLength;
 
     protected int _cfgMaxQuoteCheckChars;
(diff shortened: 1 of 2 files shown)


kdebski85:
I have just sent Contributor Agreement

cowtowncoder:
Sounds good, received CLA, hoping to review soon.

Just one request: could you rebase this against 2.13; `master` is for Jackson 3.0 which is not due out very soon?
(I just changed default branch to be `2.13` so that it is easier to know the "current" branch),

PR #283, opened 2021-08-19
getColumnDesc returns unpaired square bracket when columns are empty

pjfanning:
Returns ']' when columns list is empty - this change ensures the starting '[' is added.


cowtowncoder:
Thanks!

PR #294, opened 2021-11-07
fix one flaky test

ZeyuCai:
fix one flaky tests by using annotations. A ﬂaky test is an analysis of web application code that fails to produce the same result each time the same analysis is run. Flaky tests are detected by nondex (https://github.com/TestingResearchIllinois/NonDex) which is tool developped by the research group led by professor Darko at UIUC.
1. com.fasterxml.jackson.dataformat.yaml.misc.ObjectIdTest#testNativeSerialization


cowtowncoder:
Thanks!

Issue #288, opened 2021-09-17
Caching conflict when creating CSV schemas with different views for the same POJO


mrpiggi:
#195 added csv schema creation with views. Using the same CsvMapper with the same POJO for different views leads to unexpected results as the used view is not considered when a schema is loaded from cache as it was registered for the pojoType. Obviously, the used view should be part of the key for the cache maps `_untypedSchemas` and `_typedSchemas`.

https://github.com/FasterXML/jackson-dataformats-text/blob/1202fec8d6243da77435fc723162d738acb1c8ad/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvMapper.java#L425-L433

Edit: A workaround is quite simple. In order to clear the cache, just use `csvmapper.copy().schemaForWithView(...)` instead of `csvmapper.schemaForWithView(...)`.

cowtowncoder:
Oh. Yes indeed... Gnarly one, this.

mrpiggi:
In my eyes, caching only when `view==null` would be a sufficient fix for `2.13`

cowtowncoder:
Fixed for upcoming 2.13.1 by simply skipping caching if view enabled. May be improved in future if someone has an itch; just need to change key for cache.

mrpiggi:
What would be suitable to create a corresponding key from `JavaType pojoType` and `Class<?> view` for `LRUMap`? Is there already a generic tuple or something similar to the existing `TypedKey` integrated in `jackson-core` or `jackson-databind` that could be used for this purpose or would a new class (`ViewKey`?) be necessary?

cowtowncoder:
I can't think of a type and since we cannot add 3rd party `Pair` or `Tuple`, would need a simple key class like `ViewKey`.
@mrpiggi if you have time and interest I'd be happy to help merge this to 2.14.

`jackson-databind` probably has some similar class already fwtw, for (de)serializer caching. Not with exact types but similar.

mrpiggi:
Sure, I can try to find some time. Should I create a PR based on `2.13` or `2.14`? Are there any contribution guide lines?

cowtowncoder:
2.14: keeping caching simple for 2.13 patch just in case. 

There isn't a lot about contributing, just this:

https://github.com/FasterXML/jackson/blob/master/CONTRIBUTING.md

PR #297, opened 2021-12-17
CSV schema caching POJOs with different views

mrpiggi:
This PR adds caching of CSV schemas for POJOs with views once they have been created (see #288).

- inner class `CsvMapper.ViewKey` as cache key
- moved `SchemaCaching288Test` from `com.fasterxml.jackson.dataformat.csv.schema` to `com.fasterxml.jackson.dataformat.csv` in order to check cache usage via size

`CsvMapper.ViewKey` uses both `Objects.equals()` and `Objects.hashCode()`. Is this acceptable or should it be changed?

You would have to decide whether access modifiers for `CsvMapper.ViewKey` should be changed and whether it would be better defined as a top-level class.


cowtowncoder:
Thank you!

PR #313, opened 2022-06-25
remove use of hardcoded double value that is no longer needed

pjfanning:
The NumberInput class is copied from jackson-core and this class no longer uses this workaround - because Java 8 is minimum JDK supported.


cowtowncoder:
Thank you!

Issue #312, opened 2022-06-25
jackson-dataformats-csv: copy over recent jackson-core changes


pjfanning:
jackson-dataformats-csv has its own copy of jackson-core NumberInput and TextBuffer classes.

jackson-core changes that could be of use include:

* https://github.com/FasterXML/jackson-core/issues/577
* https://github.com/FasterXML/jackson-core/issues/514
* https://github.com/FasterXML/jackson-core/pull/757

pjfanning:
@cowtowncoder thanks for merging the fast parser/generator support - this issue also includes https://github.com/FasterXML/jackson-core/pull/757 - is that worth pursuing? That remaining issue was the one where some edge case floats are not parsed correctly because they are first parsed as doubles and then cast to float instead of being parsed explicitly as floats (2 conversions when there should just be 1).

cowtowncoder:
@pjfanning np, thank you for all these cool contributions.

As to remaining edge cases, yes, I think those would be good to address. Or at very least add (failing) tests cases where necessary. Then we could see how easy fixes are.

I also hope to get Jackson 3.0 development back on track. The one blocking thing right now is that I want to fully decide actual Java Package and Maven group ids.
My earlier idea of just doing `jackson` -> `jackson3` replacement is changing a bit; now I am leaning towards using a new domain, most likely `tools.jackson` (over `com.fasterxml`).
But once I figure out the choice there, I can finally start moving packages, so that:

1. `jackson-annotations` remains exactly as-is, as the sole exception. This means 2.x and 3.x will basically remain in-sync and there should be no need for annotation duplication (unlike with 1.x->2.0 transition)
2. `jackson-core` will move first, get published, and basically almost every other package must be modified (and re-published) by changing imports. That's lots of work and I don't want to do it many times
3. `jackson-databind` is next, similarly necessitating lots of work.
4. ... and so on for dataformats, datatypes, modules.

Oh yeah, Jackson 3.x really has no relevance to all parser work here. Just something I'll mention as relative near-term focus, possibly competing with 2.14 release. But I also do want to get 2.14.0 out quite soon -- all the must-haves I had are in (basically just the new `DataTypeFeature` -- although I think I do want to get "databind" issue 3054 implemented too).

pjfanning:
@cowtowncoder I think this can be marked as resolved now

PR #314, opened 2022-06-25
Fast parser support

pjfanning:
Relates to #312

### DIFF
--- a/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvGenerator.java
+++ b/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvGenerator.java
@@ -245,7 +245,8 @@ public class CsvGenerator extends GeneratorBase
         _ioContext = ctxt;
         _formatFeatures = csvFeatures;
         _schema = schema;
-        _writer = new CsvEncoder(ctxt, jsonFeatures, csvFeatures, out, schema);
+        boolean useFastDoubleWriter = StreamWriteFeature.USE_FAST_DOUBLE_WRITER.enabledIn(jsonFeatures);
+        _writer = new CsvEncoder(ctxt, csvFeatures, out, schema, useFastDoubleWriter);
         _writeContext = null; // just to make sure it won't be used
         _tokenWriteContext = SimpleTokenWriteContext.createRootContext(null);
         _writer.setOutputEscapes(CsvCharacterEscapes.fromCsvFeatures(csvFeatures).getEscapeCodesForAscii());
--- a/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvGenerator.java
+++ b/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvGenerator.java
@@ -245,7 +245,7 @@ public class CsvGenerator extends GeneratorBase
         _ioContext = ctxt;
         _formatFeatures = csvFeatures;
         _schema = schema;
-        _writer = new CsvEncoder(ctxt, csvFeatures, out, schema);
+        _writer = new CsvEncoder(ctxt, jsonFeatures, csvFeatures, out, schema);
         _writeContext = null; // just to make sure it won't be used
         _tokenWriteContext = SimpleTokenWriteContext.createRootContext(null);
         _writer.setOutputEscapes(CsvCharacterEscapes.fromCsvFeatures(csvFeatures).getEscapeCodesForAscii());
--- a/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/impl/TextBuffer.java
+++ b/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/impl/TextBuffer.java
@@ -313,10 +313,22 @@ public final class TextBuffer
      * Convenience method for converting contents of the buffer
      * into a Double value.
      */
-    public double contentsAsDouble()
-        throws NumberFormatException
-    {
+    public double contentsAsDouble() throws NumberFormatException {
         return NumberInput.parseDouble(contentsAsString());
+    }
+
+    /**
+     * Convenience method for converting contents of the buffer
+     * into a Double value.
+     *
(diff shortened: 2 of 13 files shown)


cowtowncoder:
Sounds good, just LMK when it's ready!

csv/src/test/java/com/fasterxml/jackson/dataformat/csv/FeaturesTest.java
```diff
@@ -26,4 +29,13 @@ public void testFactoryFeatures() throws Exception
         assertTrue(g.canUseSchema(CsvSchema.emptySchema()));
         g.close();
     }
+
+    public void testFactoryFastFeatures() throws Exception
+    {
+        CsvFactory f = new CsvFactory();
```
cowtowncoder, line 35: Maybe use builder instead, to make the test work for 3.0 too?

pjfanning: I've added an extra test for this

csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvGenerator.java
```diff
@@ -245,7 +245,7 @@ public CsvGenerator(IOContext ctxt, int jsonFeatures, int csvFeatures,
         _ioContext = ctxt;
         _formatFeatures = csvFeatures;
         _schema = schema;
-        _writer = new CsvEncoder(ctxt, csvFeatures, out, schema);
+        _writer = new CsvEncoder(ctxt, jsonFeatures, csvFeatures, out, schema);
```
cowtowncoder, line 248: Let's actually extract the new setting (boolean), pass that, not bit field. I know this may become complicated if more settings were to be passed, but for now this seems slightly cleaner.

csv/src/main/java/com/fasterxml/jackson/dataformat/csv/impl/TextBuffer.java
```diff
@@ -313,12 +313,24 @@ public BigDecimal contentsAsDecimal()
      * Convenience method for converting contents of the buffer
      * into a Double value.
      */
-    public double contentsAsDouble()
-        throws NumberFormatException
-    {
+    public double contentsAsDouble() throws NumberFormatException {
```
cowtowncoder, line 316: Mark as `@Depracated`, easier to find calls.

csv/src/test/java/com/fasterxml/jackson/dataformat/csv/deser/StreamingCSVReadTest.java
```diff
@@ -194,9 +198,9 @@ private CsvParser _parser(String csv, boolean useBytes, CsvSchema schema)
     {
         CsvParser p;
         if (useBytes) {
-            p = CSV_F.createParser(new ByteArrayInputStream(csv.getBytes("UTF-8")));
+            p = csvFactory().createParser(new ByteArrayInputStream(csv.getBytes("UTF-8")));
```
cowtowncoder, line 201: There may be a helper method `utf8Bytes(String)` in base test class (and if not can copy from `BaseTest` of `jackson-core`).

cowtowncoder:
Looks good: just one fix I'd like made wrt passing of setting.

csv/src/test/java/com/fasterxml/jackson/dataformat/csv/ser/TestGenerator.java
```diff
@@ -119,6 +120,21 @@ public void testExplicitWithFloat() throws Exception
...
 
+    public void testExplicitWithFastFloat() throws Exception
+    {
+        CsvSchema schema = CsvSchema.builder()
+                .addColumn("id")
+                .addColumn("amount")
+                .build();
+
+        float amount = 1.89f;
+        //this value loses precision when converted
```
cowtowncoder, line 131: Not sure I fully understand the comment here?

csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvGenerator.java
pjfanning: changed code to pass through a boolean

csv/src/main/java/com/fasterxml/jackson/dataformat/csv/impl/TextBuffer.java
pjfanning: done

csv/src/test/java/com/fasterxml/jackson/dataformat/csv/deser/StreamingCSVReadTest.java
pjfanning: done

csv/src/test/java/com/fasterxml/jackson/dataformat/csv/ser/TestGenerator.java
pjfanning: this new test is a copy/paste of the test above it and that pre-existing test has this comment too - I will remove the comment from both tests

cowtowncoder: SG. Let me know when all done, I'll merge this.

pjfanning:
@cowtowncoder could you re-review? I think I've addressed all the issues you highlighted.

cowtowncoder (approved):

cowtowncoder:
One potential follow-up thing: I do not recall the need for local `NumberInput` / `NumberOutput`, could perhaps remove them. In some cases duplication was, I think, in trying not to have to rely on latest `jackson-core` version (to add slightly bigger range of compatible versions). But that's not a very good reason given that there was no testing of true compatibility anyway.

For Jackson 3.0 could also consider renaming these in `jackson-core` as `JsonNumberInput` / `JsonNumberOutput` or something... not sure what'd be the best way to indicate they are focused on JSON-specific number representations which are stricter than JDK or YAML (and conceivably CSV).

PR #315, opened 2022-06-26
use jackson-core number I/O classes directly

pjfanning:
It will be harder to remove the csv module's TextBuffer because it has some extra methods that the jackson-core version doesn't.
* finishAndReturn
* looksLikeInt


cowtowncoder:
Thanks! Yeah, while it'd be nice to unify it'd probably have to be done sub-classing here. Of those two, `finishAndReturn()` might be generally useful, but `looksLikeInt()` might be less so. But I trust your assessment here.

PR #323, opened 2022-07-27
use https and add licenses to pom

pjfanning:
When changing links in pom.xml files, I noticed there were no license sections in the files. Is it ok to add them to all pom.xml files?


cowtowncoder:
I guess it's fine to add; in a way parent pom would have less duplication but maybe this is better for end users.

Could you also do PRs for `master`? Right now merging is prevented since there are always conflicts for releases (version changes).

PR #332, opened 2022-09-03
Simplest fix for issue #169 (escaping separators in JavaPropsMapper).

### LINKED ISSUE #169: No way to have properties create a map with dots in the keys
I want to create a map from properties using JavaPropsMapper, which works as long as the keys for the map don't contain dots.

I have something like this:
```
    Config expResult = new Config(
            new Config.Server(17, "/path", "param1", "param2", "Wobble")
            , ImmutableMap.<String, String>builder().put("x.y", "y").put("a.b", "b").build()
    );

    System.setProperty("server.arg", "param1");
    System.setProperty("map.x.y", "y");
```
And when I run that through the JavaPropsMapper it falls over because it creates:
```{"map":{"x":{"y":"y"}}}```
when what I want is:
```{"map":{"x.y":"y"}}```
(I want to have domain names as keys)

It would be nice if there was some way to escape the dots in property names so that they weren't considered to be delimiters.
I could probably make this work by configuring a completely different delimiter, but that would just confuse anyone wanting to work with this.

Things I've tried, any of which would be an OK solution for me, are:
```
    System.setProperty("map[x.y]", "y");
    System.setProperty("map.[x.y]", "y");
    System.setProperty("map.x\\.y", "y");
    System.setProperty("map.'x.y'", "y");
    System.setProperty("map.x'.'y", "y");
```

### PR DESCRIPTION
**Pull request by Yaytay**
This approach:
Only works for single character separators and escapes.
Only works with chars, not code points.
Removes escape chars before separator chars, leaves other escapes alone.
Sticks with the original processing until it encounters an escape before separator.

Fixes #169 

I extended MapParsingTest to test the parsing, it's got a lot of backslashes in it because they need to be doubled twice (once for Java and once for properties file).

### DIFF
--- a/properties/src/main/java/com/fasterxml/jackson/dataformat/javaprop/JavaPropsSchema.java
+++ b/properties/src/main/java/com/fasterxml/jackson/dataformat/javaprop/JavaPropsSchema.java
@@ -64,6 +64,15 @@ public class JavaPropsSchema
     protected String _pathSeparator = ".";
 
     /**
+     * Default escape character to use for single character path separators
+     * , enabling the pathSeparator to be included in a segment.
+     * Note that this is only used if the path separator is a single character
+     * and the default value (backslash) is subject to the processing of backslashes
+     * by the JDK Properties.load method (i.e. two backslashes will be required in the properties file).
+     */
+    protected char _pathSeparatorEscapeChar = '\\';
+
+    /**
      * Default start marker for index access, if any; empty String may be used
      * to indicate no marker-based index detection should be made.
      *<p>
@@ -158,6 +167,7 @@ public class JavaPropsSchema
     public JavaPropsSchema(JavaPropsSchema base) {
         _firstArrayOffset = base._firstArrayOffset;
         _pathSeparator = base._pathSeparator;
+        _pathSeparatorEscapeChar = base._pathSeparatorEscapeChar;
         _indexMarker = base._indexMarker;
         _parseSimpleIndexes = base._parseSimpleIndexes;
         _writeIndexUsingMarkers = base._writeIndexUsingMarkers;
@@ -208,6 +218,19 @@ public class JavaPropsSchema
         }
         JavaPropsSchema s = new JavaPropsSchema(this);
         s._pathSeparator = v;
+        return s;
+    }
+
+    /**
+     * Mutant factory method for constructing a new instance with
+     * specified path separator escape; default being backslash ("\").
+     */
+    public JavaPropsSchema withPathSeparatorEscapeChar(char v) {
+        if (_equals(v, _pathSeparator)) {
+            return this;
+        }
+        JavaPropsSchema s = new JavaPropsSchema(this);
+        s._pathSeparatorEscapeChar = v;
         return s;
     }
 
@@ -396,9 +419,10 @@ public class JavaPropsSchema
         return _pathSeparator;
     }
 
-    /**
-     * @since 2.10
-     */
+    public char pathSeparatorEscapeChar() {
+        return _pathSeparatorEscapeChar;
+    }
+
     public String prefix() {
         return _prefix;
     }
--- a/properties/src/main/java/com/fasterxml/jackson/dataformat/javaprop/JavaPropsSchema.java
+++ b/properties/src/main/java/com/fasterxml/jackson/dataformat/javaprop/JavaPropsSchema.java
@@ -66,9 +66,26 @@ public class JavaPropsSchema
     /**
      * Default escape character to use for single character path separators
      * , enabling the pathSeparator to be included in a segment.
-     * Note that this is only used if the path separator is a single character
-     * and the default value (backslash) is subject to the processing of backslashes
-     * by the JDK Properties.load method (i.e. two backslashes will be required in the properties file).
+     * Note that this is only used if the path separator is a single character.
+     * 
+     * The default value is NULL ('\0') which effectively disables escape processing.
+     * 
+     * The escape character is only used for escaping either the pathSeparator character
+     * or a sequence of escape characters immediately prior to the pathSeparator.
+     * i.e., if the pathSeparator is "." and the escape char is '#' then "a#.b" 
+     * produces a segment called "a.b", but "a##.b" produces a segment called "a#" 
+     * with a child called "b" and "a###.b" produces a segment called "a#.b".
+     * Finally, "a#b" produces a segment called "a#b" - the escape processing is only used
+     * immediately prior to the path separator.
+     * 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by cowtowncoder** [comment]
Ok; I'll try to get this reviewed soon now, and assuming all goes well, there's just one other thing b efore merging: I need a CLA from:

(or if you prefer, alternate Corporate CLA next to it).
That is, unless you have sent one earlier that I forgot (that has happened before :) ).
The easiest way is to print, fill & sign, scan/photo, email to `fasterxml` dot com.
Only needed once before the first contribution merge, good for all future contributions.
Apologies for the hassle but this is the only piece of process we mandate.

Looking forward to getting this merged in 2.14!

File: properties/src/main/java/com/fasterxml/jackson/dataformat/javaprop/JavaPropsSchema.java [hunk 1]
```diff
@@ -63,6 +63,15 @@
      */
     protected String _pathSeparator = ".";
 
+    /**
+     * Default escape character to use for single character path separators
+     * , enabling the pathSeparator to be included in a segment.
+     * Note that this is only used if the path separator is a single character
+     * and the default value (backslash) is subject to the processing of backslashes
+     * by the JDK Properties.load method (i.e. two backslashes will be required in the properties file).
+     */
```
cowtowncoder, line 72: Add `@since 2.14`

properties/src/main/java/com/fasterxml/jackson/dataformat/javaprop/JavaPropsSchema.java
```diff
@@ -211,6 +221,23 @@ public JavaPropsSchema withPathSeparator(String v) {
         return s;
     }
 
+    /**
+     * Mutant factory method for constructing a new instance with
+     * specified path separator escape; default being null ('\0') which
+     * should effectively disable escape processing.
+     * Any escape character may be used, backslash ('\\') is the most obvious candidate
+     * but be aware that the JDK Properties loader will dedupe all backslash characters
+     * before the JavaPropsMapper gets to see them.
```
cowtowncoder, line 230: Same here (`@since 2.14`)

properties/src/main/java/com/fasterxml/jackson/dataformat/javaprop/JavaPropsSchema.java
```diff
@@ -396,9 +423,10 @@ public String pathSeparator() {
         return _pathSeparator;
     }
 
-    /**
-     * @since 2.10
-     */
+    public char pathSeparatorEscapeChar() {
```
cowtowncoder, line 426: Let's leave `@since 2.10` next to `prefix()`, add one for 2.14 above new method.

properties/src/main/java/com/fasterxml/jackson/dataformat/javaprop/JavaPropsSchema.java
```diff
@@ -211,6 +221,23 @@ public JavaPropsSchema withPathSeparator(String v) {
         return s;
     }
 
+    /**
+     * Mutant factory method for constructing a new instance with
+     * specified path separator escape; default being null ('\0') which
+     * should effectively disable escape processing.
+     * Any escape character may be used, backslash ('\\') is the most obvious candidate
```
cowtowncoder, line 228: Based on code I think this will only work for escaping path separator character itself (at least initial index check
verifies it is followed by path separator)
That is fine but should probably be explicitly mentioned so that users do not try to use it for escaping other characters. Might be also worth mentioning that JDK properties loader will handle Unicode escapes?
(a link to explanation would be great if one can be found).

properties/src/main/java/com/fasterxml/jackson/dataformat/javaprop/util/JPropPathSplitter.java
```diff
@@ -139,12 +139,14 @@ public JPropNode splitAndAdd(JPropNode parent,
     public static class CharPathOnlySplitter extends JPropPathSplitter
     {
         protected final char _pathSeparatorChar;
+        protected final char _escapeChar;
```
cowtowncoder, line 142: Bit torn here, but perhaps `_pathSeparatorEscapeChar` would be most descriptive (given it only applies to path separator char, specifically, and is not general-purpose escape char).

properties/src/main/java/com/fasterxml/jackson/dataformat/javaprop/util/JPropPathSplitter.java
```diff
@@ -157,16 +159,64 @@ public JPropNode splitAndAdd(JPropNode parent,
...
+                            start = ix + 1;
+                            escCount = 0;
+                        }
+                    } else {
+                        segment.append(key, start, ix);
+                        curr = _addSegment(curr, segment.toString());
+                        segment = new StringBuilder();
+                        start = ix + 1;
+                    }
+                } else if (escCount > 0) {
```
cowtowncoder, line 211: Can just use `} else {` and assign `escCount = 0` -- avoids one comparison.

cowtowncoder (approved):
Looks good; added couple of notes (wrt adding `@since` tags, suggestions for javadocs).
Mostly just need the CLA.

properties/src/main/java/com/fasterxml/jackson/dataformat/javaprop/JavaPropsSchema.java
Yaytay: Whoops.

properties/src/main/java/com/fasterxml/jackson/dataformat/javaprop/JavaPropsSchema.java
Yaytay: I've expanded the documentation on both the field and the method, making them both very similar.
I've also put the same content in the README.MD (hopefully I've got the markdown right).

properties/src/main/java/com/fasterxml/jackson/dataformat/javaprop/util/JPropPathSplitter.java
Yaytay: Done.

properties/src/main/java/com/fasterxml/jackson/dataformat/javaprop/util/JPropPathSplitter.java
Yaytay: Done.

cowtowncoder:
Excellent! I also got the CLA; hoping to do one more read-through and then get this all merged in.

Thank you for working through all suggestions & for submitting the patch.

Issue #244, opened 2021-02-23
Add `YAMLGenerator.Feature.ALLOW_LONG_KEYS` to allow writing keys longer than 128 characters (default)


msmsimondean:
At the moment, a field name with a length of 128 characters or longer results in Jackson's YAML generator producing YAML like this: 

``` yaml
---
? aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
: anything
```

Using 1 less character in the field name results in the usual YAML syntax.   

``` yaml
---
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa: anything
```

Apparently the two YAML styles are called `simple keys` and `complex keys`.  SnakeYAML provides an option in its `DumperOptions` to change the max field name length at which SnakeYAML will switch from a) using the simple key style to b) using the complex key style.  Here it is: 

https://github.com/asomov/snakeyaml/blob/88c727ce266533c1db50ac24fbd3a3f33f7fe5a0/src/main/java/org/yaml/snakeyaml/DumperOptions.java#L449-L463
``` java
    public int getMaxSimpleKeyLength() {
        return maxSimpleKeyLength;
    }

    /**
     * Define max key length to use simple key (without '?')
     * More info https://yaml.org/spec/1.1/#id934537
     * @param maxSimpleKeyLength - the limit after which the key gets explicit key indicator '?'
     */
    public void setMaxSimpleKeyLength(int maxSimpleKeyLength) {
        if(maxSimpleKeyLength > 1024) {
            throw new YAMLException("The simple key must not span more than 1024 stream characters. See https://yaml.org/spec/1.1/#id934537");
        }
        this.maxSimpleKeyLength = maxSimpleKeyLength;
    }
```

Please could Jackson's YAML lib provide a `YAML feature` for using a higher value for maxSimpleKeyLength.  E.g. a LONG_SIMPLE_KEYS feature that sets maxSimpleKeyLength to something like 1024.  

[...]

PR #334, opened 2022-09-12
Allow to use long keys in YAML

ShauniArima:
Adding a new feature (`USE_LONG_KEYS`) that set `maxSimpleKeyLength` to 1024 (the maximum available length). It is set to false by default to keep the default behaviour of SnakeYAML.

It answer this issue : #244.

### DIFF
--- a/yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLGenerator.java
+++ b/yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLGenerator.java
@@ -137,6 +137,16 @@ public class YAMLGenerator extends GeneratorBase
          * @since 2.9.6
          */
         USE_PLATFORM_LINE_BREAKS(false),
+
+        /**
+         * Option passed to SnakeYAML to enable usage of key longer that 128 characters.
+         * If disabled, the max key length is set to 128 characters.
+         * <p>
+         * Default value is {@code false}.
+         * 
+         * @since 2.14
+         */
+        USE_LONG_KEYS(false),
         ;
 
         protected final boolean _defaultState;
@@ -309,6 +319,10 @@ public class YAMLGenerator extends GeneratorBase
         // 14-May-2018: [dataformats-text#84] allow use of platform linefeed
         if (Feature.USE_PLATFORM_LINE_BREAKS.enabledIn(_formatFeatures)) {
             opt.setLineBreak(DumperOptions.LineBreak.getPlatformLineBreak());
+        }
+
+        if (Feature.USE_LONG_KEYS.enabledIn(_formatFeatures)) {
+            opt.setMaxSimpleKeyLength(1024);
         }
         return opt;
     }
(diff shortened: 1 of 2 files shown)


yawkat:
It seems to me like this should be configurable as a number. But not sure if it's worth the effort for adding the API for that, and if it doesn't compromise the "hide snakeyaml" idea.

ShauniArima:
In my case, I have yaml-generated keys that can be very long and cause a lot of output with `? `. So I need to have this kind of configuration. 

In the first place, I have defined this feature as a toggle so it remains simple to use it. As you said, it can be a number. But I am not sure this is really worth it. Maybe some users really know the max lengths of their keys. If this is the main case, it will be good to use a number here. I have not found much feedback about this, do we have some?

cowtowncoder:
This sounds like a good improvement. I have just one significant concern here: backwards-compatibility. Since code as-is would have hard dependency for the new method, it would strictly require use of a specific minimum version. This may not sound like a big deal but with transitive dependency challenges chances are that it would make some upgrades more difficult.

So: which version of SnakeYAML was this added in?

Specifically I am worried about failure during runtime for cases where configuration method is NOT called -- if user does try to enable long-string support it's fair to assume version is new.
But if there's simply an upgrade in jars and no attempt is made, it'd be good not to fail.
This can be, I think, achieved by separating out code for setting the feature into separate class, which is only called when specific `Feature` is changed -- and in that case, catch and re-throw class loading failure if there is one (to indicate that an old dependency version exists).

This can be tested manually with older SnakeYAML version; not sure if there's a way (maybe using Maven profiles?) to add an automated test (integration test?) 

But at any rate, unless this setting is available in relatively old version (Jackson 2.12 depends on `1.27` for example), I think it is necessary to consider compatibility aspect.

yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLGenerator.java
```diff
@@ -310,6 +320,10 @@ protected DumperOptions buildDumperOptions(int jsonFeatures, int yamlFeatures,
         if (Feature.USE_PLATFORM_LINE_BREAKS.enabledIn(_formatFeatures)) {
             opt.setLineBreak(DumperOptions.LineBreak.getPlatformLineBreak());
         }
+
+        if (Feature.USE_LONG_KEYS.enabledIn(_formatFeatures)) {
+            opt.setMaxSimpleKeyLength(1024);
```
cowtowncoder, line 325: This is the setting that would need to be encapsulated in a separate class, with a method that gets `opt` passed in, sets the maximum length. Since that class is only loaded when method called (first "active use") it is possible
to avoid failure when `YAMLGenerator` class itself is initialized.
Class loading exception should trigger throwing of exception that suggests too old SnakeYAML dependency.

ShauniArima:
According to [this ticket](https://bitbucket.org/snakeyaml/snakeyaml/issues/431/increase-simple-key-length-to-1024-stream). This feature has been available since version 1.24 of SnakeYaml.

I will work on adding a wrapper to make it less error-prone as you explain it. 😄

cowtowncoder:
@ShauniArima Actually if it's that old a fix, never mind -- Jackson 2.10.0 already had snakeyaml 1.24 that as the dependency version. I am not as worried in that case, wrt 2.14.0.

ShauniArima:
I have filled and signed the CLA. You will receive a mail soon. 😉

Issue #337, opened 2022-09-19
Allow overriding of file size limit for YAMLParser by exposing SnakeYAML `LoaderOptions`


pjfanning:
Relates to https://github.com/FasterXML/jackson-dataformats-text/pull/335#issuecomment-1251631287

snakeyaml 1.32 brings in a default limit of 3Mb when parsing yaml files.

Need to allow users to specify another value if they need to.

https://bitbucket.org/snakeyaml/snakeyaml/src/72dfa9f1074abe2b8a6c8776bee4476b0aed02e3/src/main/java/org/yaml/snakeyaml/LoaderOptions.java

cowtowncoder:
See my notes on #335: there will be need, likely, to carefully apply any new configurations so as to avoid having a hard limit on very recent SnakeYAML versions, assuming configurability will be added post-1.32 (which seems likely).

pjfanning:
@cowtowncoder probably best to revert the v2.13 branch to stick with snakeyaml 1.31 - with v2.14, I see YAMLParser has feature support but this is boolean - is there a preferred to set int properties?

cowtowncoder:
@pjfanning That might make sense (2.13).

As to `int` value configuration, right, I think it has to go via `YAMLFactory.Builder` and no there isn't anything out of the box. I guess there's no point in doing on/off setting here, it being a security feature and realistically I think it makes sense to always have some upper limit, even it was `Integer.MAX_VALUE`.

On Builder, further, it may get bit trickier with 2.x vs 3.0 since 2.x has no immutability requirement but 3.0 has.

cowtowncoder:
Added notes to the PR: my position now is that I kind of like exposing `LoaderOptions`, to make users configure aspects they want to. This reduces some of compatibility problems, and increases configurability significantly.

It will also be a problem for 2.x/3.0 migration, likely; may need to see how snakeyaml-engine deals with this. Seems likely there is a counterpart to this configuration object, and if so, it may be relatively easy to offer similar extension point.

Put another way, I am ok with tighter bundling of SnakeYAML at least for Jackson 2.x.
There are no alternatives that I know of, and really my only wish would be performance improvements. But functionality is solid and author has been very quick with fixes, releases.

pjfanning:
@cowtowncoder LoaderOptions seems to have been added to snakeyaml in v1.18

cowtowncoder:
[...]

PR #339, opened 2022-09-20
support overriding the codePointLimit

pjfanning:
https://github.com/FasterXML/jackson-dataformats-text/issues/337

Work in progress. Will add tests before merging but would like to open a discussion on the APIs.

Another option is to support setting a snakeyaml LoaderOptions instance on YamlFactoryBuilder, so users would have full control. There are other settings in LoaderOptions that might be useful to support. https://www.javadoc.io/doc/org.yaml/snakeyaml/latest/org/yaml/snakeyaml/LoaderOptions.html

### DIFF
--- a/yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLFactory.java
+++ b/yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLFactory.java
@@ -10,6 +10,7 @@ import com.fasterxml.jackson.core.format.InputAccessor;
 import com.fasterxml.jackson.core.format.MatchStrength;
 import com.fasterxml.jackson.core.io.IOContext;
 import com.fasterxml.jackson.dataformat.yaml.util.StringQuotingChecker;
+import org.yaml.snakeyaml.LoaderOptions;
 
 @SuppressWarnings("resource")
 public class YAMLFactory extends JsonFactory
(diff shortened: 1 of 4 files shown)


cowtowncoder:
I think that although in many ways it's good to hide details of the underlying implementation, in this case benefits probably outweigh drawbacks.

One particularly compelling reason here is this: as I understand, this new setting is only available in SnakeYAML 1.32. If so, we either have to:

1. Require baseline of SnakeYAML from users. This will almost certainly gain as many bug reports for breakages from users with different version of SnakeYAML in their classpath
2. Add dynamic loading of piece of code that handles with this setting, to only fail if someone tries to set the limit against older SnakeYAML

But if we only expose functionality for interacting with `LoaderOptions` (when was it added?), we can avoid this problem altogether.

If it wasn't for this compatibility issue, I might suggest doing both (simple `int` setting and `LoaderOptions`), but due to it I suggest exposing `LoaderOptions`. This assumes it was added in a version that is not "too recent" (like, something that was the baseline for... maybe latest Jackson 2.9.x release?).

pjfanning:
@cowtowncoder I have reworked to this to allow users to set the LoaderOptions instance explicitly

cowtowncoder:
Looks good, will merge.

One thing that might make sense -- and I can try to add it -- would be an example of how to expand maximum size limit, on  YAML module's README.

cowtowncoder:
@pjfanning Was able to merge to `master`, but it looks as if `snakeyaml-engine` did not have equivalent setting: while `LoaderOptions` became `LoadSettings`, I did not see an equivalent to "max code points" in latter.

@asomov Do you know if I missed something obvious (setting is maybe somewhere else), or if omission is intentional?

asomov:
@cowtowncoder it does not have it yet. But it will get it soon

cowtowncoder:
@asomov Excellent: no super hurry as this only affects Jackson 3.0.0-SNAPSHOT, not yet released. So just let us know when it is added. Thank you!

asomov:
@cowtowncoder SnakeYAML Engine version 2.5 with the support for codePointLimit [is released](https://search.maven.org/artifact/org.snakeyaml/snakeyaml-engine/2.5/bundle)

cowtowncoder:
Thank you @asomov ! I enabled test for this in `master` for JAckson YAML module against `snakeyaml-engine` too.

/cc @pjfanning

PR #341, opened 2022-09-23
use NumberInput.parseBigInteger

pjfanning:


cowtowncoder:
Hmmmh. I guess it makes sense, although it also limits compatibility of "adjacent" versions -- textual dataformats now definitely require `jackson-core` of 2.14.

PR #356, opened 2022-11-22
Fix TOML parse failure when number token hits buffer edge

yawkat:
When a number token is exactly at the end of the lexer text buffer, the parser would advance the lexer, triggering a buffer refill, before the number is parsed from the buffer. This patch moves the advance operation to come after parsing, which resolves the issue.

Kudos to @wbprime for finding this and reporting it as https://github.com/micronaut-projects/micronaut-toml/issues/93 . Not even the fuzzer managed to hit this – maybe because it does not fail fast, and just returns the wrong output (NumberInput does no input checking)?


yawkat:
hm outstanding test failures...

yawkat:
i think the fuzz tests just fail because of illegal tokens now, but as long as there's a failure, this seems fine

cowtowncoder:
@yawkat Correct, many/most methods in `NumberInput` expect caller to pass valid Strings (since that's what original JSON parser does before calls anyway). This is good for performance but can in fact hide errors like buffer boundary ones.
Fuzzer also has the challenge that it can find buffer boundary if it results ArrayIndexOutOfBoundsException but not if it just produces invalid output (like you suggested).

cowtowncoder (approved):

PR #357, opened 2022-12-03
Fast Big decimal/integer parser

pjfanning:
Had to make some changes to the TOML module - to open up the parser to allow the StreamReadFeatures to be set.


cowtowncoder (approved):

cowtowncoder:
Yeah... `TomlFactory` really should not work that way from static methods. I should take time to see if it can be changed.

cowtowncoder:
Ah. Ok, so those odd `static` parser methods in `Parser` (which is not a `JsonParser`) are... hmmh.
Those really should be untangled somehow: I think that actual `TomlParser` should be created even if it uses `TreeTraversingParser` under the hood. This would be needed to properly expose features etc.

But that's for another time; I just hope my merge didn't break anything too obvious.

cowtowncoder:
Went ahead and cleaned things up, reduced number of places where static `Parser.parse()` is needed to just couple.
Ideally would not need by tests but there's one odd test constant that prevents this for now.

PR #362, opened 2022-12-15
Add max-length checks into YAMLParser (similar to JsonParserBase)

cowtowncoder:


cowtowncoder:
@pjfanning This was missing; hope I didn't miss obvious things. Exception does change but I think it is probably more correct for 3.0 to just expose "low-level" `StreamReadException`. At some point probably need to unify it but right now just want to catch the issue.

pjfanning:
@cowtowncoder could this be backported to v2.15 branch?

cowtowncoder:
@pjfanning Hmmh. I was somehow assuming 2.15 just worked (as no failing tests). But makes sense it wouldn't as check is just not there.

As to running snapshot, it should have automatically done that shouldn't it? Or, rather, my earlier publish should have synced I don't understand why snapshot wouldn't be there -- I did `./mvnw clean deploy` and that does push a new snapshot.
I mean I can re-run but I don't think there's old version remaining.

cowtowncoder:
@pjfanning invited you to "core" team that should give necessary access.

cowtowncoder:
@pjfanning I also re-ran databind master build, which passed. None of test runs for `jackson-dataformats-text` (for 2.15 or master) fail so I cannot really reproduce the issues, somehow. I will try to backport this change but it is peculiar that I cannot reproduce problem(s) you have.

pjfanning:
> @pjfanning invited you to "core" team that should give necessary access.
> 
> I also re-did snapshot release from local system for `jackson-databind` 2.15.0-SNAPSHOT and 3.0.0-SNAPSHOT.

If you manually pushed a new snapshot, then that's ok.

The tests passed for me on master when I ran on my laptop yesterday. So I'm quite confused as to why these tests were failing and why this PR was needed.

We now have a minor diff between v2.15 and v3 in that different exceptions are thrown.

cowtowncoder:
Ok yes then we are synced. I am now backporting changes but not 100% sure why they weren't exactly needed.

Discrepancy is unfortunate but for now I think I don't have time to dig into it deeper.

PR #363, opened 2022-12-18
csv deserializer needs to apply num len checks

pjfanning:
* try to ensure CSV deserializer code checks num lengths in more places
* also one place in YAML code where custom number parsing happens

### DIFF
--- a/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/impl/CsvDecoder.java
+++ b/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/impl/CsvDecoder.java
@@ -1151,6 +1151,7 @@ public class CsvDecoder
         } else if (_numberString == null) {
             throw new IllegalStateException("cannot get BigInteger from current parser state");
         }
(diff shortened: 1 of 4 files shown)


pjfanning:
@cowtowncoder would it be possible to review this PR? It covers extra cases where NumberInput parse number calls are made that have not been previously covered.

csv/src/main/java/com/fasterxml/jackson/dataformat/csv/impl/CsvTextBuffer.java
```diff
@@ -99,8 +102,16 @@
     /**********************************************************
      */
 
+    @Deprecated // @since 2.15
     public CsvTextBuffer(BufferRecycler allocator)
     {
+        _streamReadConstraints = null;
```
cowtowncoder, line 108: I'll change this to just use default one, to avoid null checks.

cowtowncoder:
> @cowtowncoder would it be possible to review this PR? It covers extra cases where NumberInput parse number calls are made that have not been previously covered.

Yes, going slowly through my list of backlog. Will merge now.

PR #365, opened 2022-12-19
Rename local `TextBuffer` as `CsvTextBuffer` to make customization explicit

cowtowncoder:


cowtowncoder:
cc @pjfanning Will rename so it's easier to see we currently used a modified copy. I am trying to remember why this was done; there's `CsvIOContext` too. I think that ideally we would try to use standard `TextBuffer` instead but am drawing blank on what lead to copy-pasting here.

PR #369, opened 2023-01-10
Do not use deprecated code in SnakeYAML

asomov:
Deprecated code in SnakeYAML will be removed in the coming release


cowtowncoder:
Created #371 for 2.14 instead; will use that. Thank you, @asomov !

PR #377, opened 2023-01-31
fix broken yaml tests - implement getNumberValueDeferred

pjfanning:
* relates to https://github.com/FasterXML/jackson-databind/pull/3761

With YAMLParser, this version is too eager but makes all the tests pass:

    @Override
    public Object getNumberValueDeferred() throws IOException {
        return getDecimalValue();
    }
YAMLParser inherits the getNumberValueDeferred from ParserBase and that has a relatively complicated implementation that doesn't seem to work well for YAMLParser.

`return getNumberValue();` leads to small rounding issues that affect tests

### DIFF
--- a/yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLParser.java
+++ b/yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLParser.java
@@ -992,6 +992,11 @@ public class YAMLParser extends ParserBase
     /* Number accessor overrides
     /**********************************************************************
      */
+
+    @Override
+    public Object getNumberValueDeferred() throws IOException {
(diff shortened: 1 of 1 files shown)


cowtowncoder:
Hmmh. This does not look right. As in: there seems to be some bigger issue at play here; possibly with YAML parser (that is, Jackson `YAMLParser`, not SnakeYAML). For some reason, empty String gets buffered...

cowtowncoder:
Ok; I decided to look deeper into this. Not 100% sure why breakage occurred; I think it is due to complications with lazy/deferred Double/Float parsing, something that didn't originally break anything but in combination with other changes now does.

But I was able to write bit more optimal implementation of `getNumberValueDeferred()` which passes, so closing this one.

PR #389, opened 2023-02-24
feat: add feature flag to parse boolean like words as strings

### LINKED ISSUE #388: Add feature flag to allow parsing "boolean" words as strings instead of booleans
Right now [YAMLParser._matchYAMLBoolean](https://github.com/FasterXML/jackson-dataformats-text/blob/2.15/yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLParser.java#L637) provides no flexibility to avoid words that are considered booleans to be parsed as strings instead, which is actually sometimes the desired behaviour and a good candidate for a feature flag in my opinion. I therefore propose the following

- Introduce a new `YAMLParser.Feature` called `PARSE_TRUTHY_WORDS_AS_STRINGS` (open for suggestion on names)
- When enabling this flag, the following words (converted to lowercase) will be parsed as strings instead of booleans:
  - yes
  - no
  - on
  - off
  - y
  - n

Note that `false` and `true` will still be parsed as booleans.

Let me know what you think! I'll be happy to help out with a PR here.

Related issue: https://github.com/FasterXML/jackson-dataformats-text/issues/129

axelniklasson:
Introduce a new feature flag to `YAMLParser` that when explicitly enabled, leads to the following words (irrespective of case) being parsed as strings instead of booleans:

- `yes`
- `no`
- `off`
- `on`

Note that `y`/`n` are already parsed as strings and this change will not impact how `true` and `false` are being parsed - they will still be parsed as booleans.

Fixes #388

### DIFF
--- a/yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLParser.java
+++ b/yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLParser.java
@@ -44,6 +44,8 @@ public class YAMLParser extends ParserBase
         /**
          * Feature that determines whether to parse boolean-like words as strings instead of booleans.
          * When enabled, the following words will be parsed as strings instead of booleans: yes, no, on, off.
+         * 
+         * @since 2.15
          */
         PARSE_BOOLEAN_LIKE_WORDS_AS_STRINGS(false),
         ;
--- a/yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLParser.java
+++ b/yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLParser.java
@@ -39,7 +39,13 @@ public class YAMLParser extends ParserBase
          * Feature is enabled by default in Jackson 2.12 for backwards-compatibility
          * reasons.
          */
-        EMPTY_STRING_AS_NULL(true)
+        EMPTY_STRING_AS_NULL(true),
+
+        /**
+         * Feature that determines whether to parse boolean-like words as strings instead of booleans.
+         * When enabled, the following words will be parsed as strings instead of booleans: yes, no, on, off.
+         */
+        PARSE_BOOLEAN_LIKE_WORDS_AS_STRINGS(false),
         ;
 
         final boolean _defaultState;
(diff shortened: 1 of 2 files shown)


yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLParser.java
```diff
@@ -39,7 +39,13 @@ public enum Feature implements FormatFeature // in 2.9
          * Feature is enabled by default in Jackson 2.12 for backwards-compatibility
          * reasons.
          */
-        EMPTY_STRING_AS_NULL(true)
+        EMPTY_STRING_AS_NULL(true),
+
+        /**
+         * Feature that determines whether to parse boolean-like words as strings instead of booleans.
+         * When enabled, the following words will be parsed as strings instead of booleans: yes, no, on, off.
+         */
```
cowtowncoder, line 47: please add `@since 2.15).

cowtowncoder:
Ok I think this looks fine on first look (I'll need to go over it once more before merging just in case but nothing off so far). So I think I'd like to get it merged.

But one (and only) process thing: if I haven't asked for and gotten CLA:

yet, it'd be needed before the first contribution (and it's then good for any number of contributions to Jackson project).

The usual way is to print it, fill & sign, scan/photo, email to `info` at fasterxml dot com.

Once I get that I can proceed with merging.

Thank you for this contribution & looking forward to merging it!

axelniklasson:
Thanks for looking @cowtowncoder! Added the `@since` tag and sent over the signed CLA as well.

cowtowncoder:
CLA received, doing last review now.

yaml/src/test/java/com/fasterxml/jackson/dataformat/yaml/deser/ParseBooleanLikeWordsAsStringsTest.java
```diff
@@ -0,0 +1,85 @@
...
+        assertEquals(root.get("seven").getNodeType(), JsonNodeType.STRING);
+        assertEquals(root.get("seven").textValue(), "Y");
+
+        assertEquals(root.get("eight").getNodeType(), JsonNodeType.STRING);
+        assertEquals(root.get("eight").textValue(), "N");
+    }
+
+    public void testParseBooleanLikeWordsAsString_enabledFF() throws Exception
+    {
+        YAMLFactory f = new YAMLFactory().enable(YAMLParser.Feature.PARSE_BOOLEAN_LIKE_WORDS_AS_STRINGS);
```
cowtowncoder, line 56: I'll change this to use builder-based configuration, merges more cleanly in `master` (3.0 has immutable factories).

cowtowncoder (approved):

axelniklasson:
Thanks for the quick turnaround @cowtowncoder, happy to see this merged! Do you by any chance have a rough timeline for when 2.15 might be available?

cowtowncoder:
@axelniklasson Hoping to get release candidate out in early March (in 2 weeks?), and from that, 3-4 weeks for final release.

PR #390, opened 2023-02-26
Upgrade to Snakeyaml 2.0 (resolves CVE-2022-1471)

pjfanning:
CVE fix: https://bitbucket.org/snakeyaml/snakeyaml/issues/561/cve-2022-1471-vulnerability-in

* tests pass alright
* may be a bit risky to put into jackson 2.14 (snakeyaml 2.0 is a major release)

### DIFF
--- a/yaml/pom.xml
+++ b/yaml/pom.xml
@@ -34,7 +34,7 @@
     <dependency>
       <groupId>org.yaml</groupId>
       <artifactId>snakeyaml</artifactId>
-      <version>1.33</version>
+      <version>2.0</version>
     </dependency>
 
      <!-- and for testing need annotations; but should be available via `jackson-databind` above


cowtowncoder:
Hmmmh. Might be bit risky even for 2.15, no?

I am guessing tho that major upgrade here is not due to API change but for... marketing?
Do you have more info on this?

pjfanning:
Snakeyaml was refactored a bit to make it more secure by default. Refactor doesn't seem to break anything Jackson wise. It does affect other code bases that use more of the snakeyaml APIs 

My view is that Jackson upgrading is probably ok. And means we avoid the CVE marker on our transitive dependencies. Users who use this module and another snakeyaml dependent lib may run into issues with the other lib not yet up taking the latest snakeyaml.

cowtowncoder:
Ah. And 2.0 version bump is not for marketing but due to defaults change which is behavioral.
And since SnakeYAML does not use patch numbers -- that is, only uses 2 digits, it is not "small enough" to qualify for `1.34` -- it makes sense.

I'll merge this and then need to add some verbiage in 2.15 release notes: I agree it's unlikely this should cause significant breakage compared to, say, earlier introduction of processing limits wrt input doc size.

Thank you for staying on top of this @pjfanning: those transitive dep CVEs are unfortunate for users.

cowtowncoder:
Merged, added a note on 

https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.15

pjfanning:
> Merged, added a note on
> 
> https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.15

The release note says ` upgrade to Snakeyaml 2.0 from 2.13` - I'm not sure what the 2.13 means.

One thing that I've run into is that Spring framework seems to use snakeyaml to read config files and the Spring framework code seems to use APIs that break with Snakeyaml 2.0. See the test failures in https://pipelines.actions.githubusercontent.com/serviceHosts/2b2c8c25-f923-4d14-bc54-07f0604e4241/_apis/pipelines/1/runs/19578/signedlogcontent/2?urlExpires=2023-02-26T14%3A37%3A10.9289750Z&urlSigningMethod=HMACV1&urlSignature=sV%2ByS%2BFcZ9mx2OAGZDj9jvplw%2BRYYwVAHctlPnHzBRU%3D

I still think that anyone who uses a lib that breaks with snakeyaml 2.0 can choose to hold off the Jackson upgrade until the other lib is fixed up. Everyone else benefits from using the more secure snakeyaml release.

cowtowncoder:
Ok that was my fat fingering; was to be to 1.33. Thanks, will fix.

Hmmmh. Spring breakage, if it cannot be solved from their side, is something that would make me consider reverting this upgrade -- considering that one can override snakeyaml dependency for Jackson for newer version easily.
I guess technically one can also downgrade it but that has the issue of CVE reporting.

So I guess we need to keep this in mind for 2.15.0 release candidate phase and see what makes most sense.

pjfanning:
I looked a bit closer at the Apache Linkis issue and they are using a fairly old bundle of Spring jars.

It seems like newer Spring versions may be better off. I looked at the Spring code and it has been adjusted to work with a newer 1.x version of snakeyaml.

It looks like some deprecated methods and constructors were removed in snakeyaml 2.0.

cowtowncoder:
@pjfanning Ok thank you. I think we can deal with this as part of pre-release cycle then.

ns-jhu:
This is available now:

```


<         <fasterxml.jackson-core.version>2.14.2</fasterxml.jackson-core.version>
<         <fasterxml.jackson-databind.version>2.14.2</fasterxml.jackson-databind.version>
<         <fasterxml.jackson-dataformat.version>2.14.2</fasterxml.jackson-dataformat.version>
<         <fasterxml.jackson-annotations.version>2.14.2</fasterxml.jackson-annotations.version>
---
>         <fasterxml.jackson-core.version>2.15.0-rc2</fasterxml.jackson-core.version>
>         <fasterxml.jackson-databind.version>2.15.0-rc2</fasterxml.jackson-databind.version>
>         <fasterxml.jackson-dataformat.version>2.15.0-rc2</fasterxml.jackson-dataformat.version>
>         <fasterxml.jackson-annotations.version>2.15.0-rc2</fasterxml.jackson-annotations.version>




```
and I confirmed that this works  too.

PR #391, opened 2023-02-26
remove unnecessary code in YamlParser

pjfanning:
My IDE shows these lines are being unnecessary. `continue` just skips to the end of the loop and the loop starts again. And these lines are already at the end of the loop. Removing this code might improve performance a little bit.


cowtowncoder:
Ok sure. I think what is missing is an error for anything unrecognized, without which this check makes no difference.
I doubt it has measurable effect on performance (YAML decoder is VERY slow all around) but there's no point in leaving dead code so will merge.

PR #393, opened 2023-03-04
Constraints issues

pjfanning:
after recent jackson-core changes

### DIFF
--- a/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvParser.java
+++ b/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvParser.java
@@ -658,6 +658,10 @@ public class CsvParser
         return false;
     }
 
+    /**
+     * @return whether the token is expected to be an IntToken
+     * @throws IllegalStateException if the number is too long
+     */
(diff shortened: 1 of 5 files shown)


csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvParser.java
```diff
@@ -671,7 +675,7 @@ public boolean isExpectedNumberIntToken()
                     return true;
                 }
             } catch (IOException e) {
-                throw new RuntimeException(e);
+                throw new IllegalStateException(e);
```
pjfanning, line 678: @cowtowncoder you've applied most of the changes that were in this PR - but this is the one extra bit: would it be possible to use IllegalStateException or UncheckedIOException here?

cowtowncoder: Ah, yes, thank you for reminding me -- I think I'll go with `UncheckedIOException` as in another place.
I had vague recollection there was one more, this was it.

cowtowncoder (approved):
Will merge, modify

PR #394, opened 2023-03-05
[DRAFT] validate deep nesting

pjfanning:
relies on https://github.com/FasterXML/jackson-core/pull/943

### DIFF
--- a/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvParser.java
+++ b/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvParser.java
@@ -626,6 +626,8 @@ public class CsvParser
      * We need to override this method to support coercion from basic
      * String value into array, in cases where schema does not
      * specify actual type.
+     *
+     * @throws UncheckedIOException if the nesting depth is too large, see {@link StreamReadConstraints.Builder#maxNestingDepth(int)}
      */
     @Override
     public boolean isExpectedStartArrayToken() {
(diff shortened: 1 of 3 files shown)


csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvParser.java
```diff
@@ -1403,11 +1409,16 @@ public ByteArrayBuilder _getByteArrayBuilder()
         return _byteArrayBuilder;
     }
 
-    protected void _startArray(CsvSchema.Column column)
+    protected void _startArray(CsvSchema.Column column) throws UncheckedIOException
```
cowtowncoder, line 1412: Is there a reason for this not to expose `IOException`?

pjfanning: changed just now - I'm not sure if this PR is really useful though, I can't really see how you can nest CSV or JavaProperties - I'm going to concentrate on use cases like Smile and CBOR first - YAML is covered by SnakeYAML LoaderOPtions (so like jackson-dataformat-xml, the argument is if we need 2 ways to police the depth, SnakeYAML has a default depth limit of 50).

cowtowncoder: Ah. Yeah, good point. In theory in future we _might_ create nested Objects etc. But even then... seems unlikely & can cross that bridge if we get there.

For Properties it's bit different since there is the path splitting so one can actually create keys to create whatever depth.

And YAML is interesting as limits are needed but... provided natively. So can/should postpone handling changes there too I think.

TOML, once again is... its own thing.

So I think Properties are one to consider, yet less important than Smile, CBOR. Can't do it for Ion I suspect.
Protobuf doesn't have constructs for infinite recursion I don't think?

pjfanning:
closing in favour of #397

PR #396, opened 2023-03-06
add LoaderOptions doc to README

pjfanning:


yaml/README.md
```diff
@@ -60,7 +60,16 @@ while (parser.nextToken() != null) {
 Most configuration is applied during mapper instance configuration, through
 `YAMLMapper.Builder`, similar to how JSON-based plain `ObjectMapper` is configured.
 
-## Known problems
+If you are parsing YAML, it is recommended that you consider configuring the
```
cowtowncoder, line 63: I think this must note version this applicable to (2.14+), otherwise we get questions on why settings are missing.

Also I am not 100% sure it is really recommended in the sense that the defaults should be fine for most users?
Maybe change wording to suggest it may be necessary when processing larger content or something.
I do not think it is something that really affects vast majority of users.

pjfanning: I refactored the doc to remove the word 'recommend' and also highlighted that this is about v2.14 and above

cowtowncoder (approved):
Perfect. Thank you again!

PR #399, opened 2023-03-07
add toml tests for oss-fuzz reports

pjfanning:
* recent nestingDepth changes fix 2 open oss-fuzz issues - 1 was tested in last commit but added the other for completeness
* add IndexOutOfBoundException catch to fix the 3rd


cowtowncoder (approved):

toml/src/main/java/com/fasterxml/jackson/dataformat/toml/Parser.java
```diff
@@ -78,15 +78,20 @@ public static ObjectNode parse(
...
+                    factory.getFormatParserFeatures(), reader);
+            try {
+                final ObjectNode node = parser.parse();
+                assert parser.getNestingDepth() == 0;
+                return node;
+            } finally {
+                parser.lexer.releaseBuffers();
+            }
+        } catch (IndexOutOfBoundsException indexOutOfBoundsException) {
+            // https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=51654
```
cowtowncoder, line 93: This seems like a bug in lexer? I assume input is invalid but not sure if there was a chance to fix it there (and are we hiding it here).

pjfanning: Lexer.java is generated from Lexer.jflex and the exception occurs specifically in code that is auto-generated (because some of the generated code in Lexer.java does come verbatim from the jflex file). I am no jflex expert but it would appear that we can't do much more than just catch the IndexOutOfBoundsException.

yawkat: could maybe move the try-catch around the yylex calls instead

pjfanning: I created https://github.com/FasterXML/jackson-dataformats-text/pull/402

PR #402, opened 2023-03-08
move try-catch for IndexOutOfBoundsException

pjfanning:


yawkat:
give me a minute, i think it's possible to change the skeleton to fix the exception happening at all

cowtowncoder:
I think this is no longer needed after merging #403?

PR #403, opened 2023-03-08
toml: Ensure progress in zzRefill

yawkat:
The IndexOutOfBoundsException is caused by a condition where the buffer is still large enough that it can hold one char, but that char is a high surrogate, so zzRefill returns false (no EOF) but does not increase zzEndRead (the high surrogate does not have a corresponding low surrogate yet).

This change simply repeats the zzRefill code until at least one character is fully available.

See #399 #402

### DIFF
--- a/toml/pom.xml
+++ b/toml/pom.xml
@@ -65,7 +65,7 @@
             <plugin>
                 <groupId>de.jflex</groupId>
                 <artifactId>jflex-maven-plugin</artifactId>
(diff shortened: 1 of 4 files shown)


yawkat:
this bug should also happen with the default jflex skeleton, but unfortunately i dont understand the jflex test system so i dont know how to submit a patch there...

cowtowncoder (approved):

Issue #413, opened 2023-04-12
SnakeYAML LoaderOptions.setAllowDuplicateKeys(false) should fail on duplicate keys.


nielsbasjes:
When creating a YAMLFactory we can specify the SnakeYAML `LoaderOptions`.

One of the settings in there is to make the load fail if a duplicate key is found in the parsed yaml.

This feature does not work when using Jackson.

To demonstrate (I'm donating these tests to your project) two ways of parsing the same document
- the first test  (using the SnakeYaml code) passes (i.e. throws a DuplicateKeyException)
- the second test (using the Jackson code) fails (i.e. does NOT throw a DuplicateKeyException)

```
String yamlDocument = "key1: a\nkey1: b";

@Test
void loadingDuplicateKeysShouldFailIfNotAllowedSnakeYaml() {
    LoaderOptions yamlLoaderOptions = new LoaderOptions();
    yamlLoaderOptions.setAllowDuplicateKeys(false);

    Yaml yaml = new Yaml(yamlLoaderOptions);
    assertThrows(DuplicateKeyException.class, () -> {
        yaml.load(yamlDocument);
    });
}

@Test
void loadingDuplicateKeysShouldFailIfNotAllowedJackson() throws IOException {
    LoaderOptions yamlLoaderOptions = new LoaderOptions();
    yamlLoaderOptions.setAllowDuplicateKeys(false);

    YAMLFactory yamlFactory = YAMLFactory.builder().loaderOptions(yamlLoaderOptions).build();

    try(YAMLParser yamlParser = yamlFactory.createParser(yamlDocument)) {
        assertThrows(DuplicateKeyException.class, () -> {
            new ObjectMapper().readTree(yamlParser);
        });
    }
}
```

yawkat:
only the options supported by snakeyaml's ParserImpl are supported by jackson. allowDuplicateKeys is an option of the snakeyaml binder, not the parser, so it's not supported.

imo the current behavior is reasonable.

nielsbasjes:
I find being able to fail on duplicate keys a feature I enable most of the time.

Right now I'm setting a flag that is supported by the `LoaderOptions` yet it does not have the documented effect because the way it has currently been implemented.

I understand that if you have implemented it in a way that cannot support this, but then I should get some kind of "unsupported feature" exception when I try to do this.

Also is there an other way achieve the same effect (i.e. fail on duplicate keys)?

cowtowncoder:
[...]

PR #415, opened 2023-04-13
Use `LoaderOptions.allowDuplicateKeys` to enforce duplicate key detection

nielsbasjes:
If the user sets the allowDuplicateKeys to false in the LoaderOptions they expect the system will fail on a duplicate key.
It does not (see https://github.com/FasterXML/jackson-dataformats-text/issues/413 ) and no errors are given to indicate they should have set it in a different way.

With this simple change it will fail on duplicate keys.


cowtowncoder (approved):

cowtowncoder:
Looks good. Just one piece of paperwork: if you haven't yet sent CLA (just needs to be done once, ever), we'd need this:

(and usually it's easiest to print, fill & sign, scan/photo, email to `info` at fasterxml dot com)

and once we have it, can proceed merging the PR.

Thank you for the PR; I think this is a nice ergonomic improvement!

nielsbasjes:
@cowtowncoder I have sent the signed CLA just now.

cowtowncoder:
CLA received.

PR #418, opened 2023-05-01
Fix #386 Option to add .ignorePropery() to Csv builder

### LINKED ISSUE #386: Add .ignorePropery() to csv builder
I would love to add fieldnames that should be ignored, like:

`mapper.schemaFor(ExternalEntity.class).withHeader().ignoreProperty('id');`

I know the `JsonIgnoreProperty` annotation, but that requires modifying the target class, which is not always possible.
Thus, a configuration setter would be required.

djay-S:
Added new method CsvSchema.ignoreProperty() to define the properties that need to be removed for the csv.
Issue: https://github.com/FasterXML/jackson-dataformats-text/issues/386

### DIFF
--- a/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvSchema.java
+++ b/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvSchema.java
@@ -567,6 +567,15 @@ public class CsvSchema
         }
 
         /**
+         * Helper method called to remove a column
+         *
+         * @since 2.17.2
+         */
+        public boolean removeColumn(Column column) {
+            return _columns.remove(column);
+        }
+
+        /**
          * Helper method called to drop the last collected column name if
          * it is empty: called if {link CsvParser.Feature#ALLOW_TRAILING_COMMA}
          * enabled to remove the last entry after being added initially.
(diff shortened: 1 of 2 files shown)


djay-S:
Updated the feature branch with the latest code corrections from the base branch.

cowtowncoder:
@djay-S I may be wrong, but I suspect this does not quite do what #386 is asking: I _think_ it asks for something that would work similar to how `@JsonIgnore` works on POJO, but defined via schema. This PR on the other hand simply removes column definition which has other side effects.

But I hope @membersound can clarify details of how he thinks feature should work.

djay-S:
@cowtowncoder I have made changes for not removing the column definitions and only ignoring the columns. Let me know if the changes are valid.

membersound:
> @djay-S I may be wrong, but I suspect this does not quite do what #386 is asking: I _think_ it asks for something that would work similar to how `@JsonIgnore` works on POJO, but defined via schema. This PR on the other hand simply removes column definition which has other side effects.
> 
> But I hope @membersound can clarify details of how he thinks feature should work.

Exactly: I simply need a method that behaves like `@JsonIgnore`, but without having to tough the target bean codewise.

csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvSchema.java
```diff
@@ -1483,6 +1483,22 @@ public String getColumnDesc()
         sb.append(']');
         return sb.toString();
     }
+    /**
+     * Method for ignoring fields for CsvBuilder
+     * @param ignoreProperties Array of column names to be ignored by the csv builder
+     * @return A newly built {@link CsvSchema} with the ignored properties
+     */
+    public CsvSchema ignoreProperty(String[] ignoreProperties) {
```
cowtowncoder, line 1491: From naming perspective it should probably be "ignoreProperties".

Also, would perhaps make sense to take `String...` instead of `String[]` for convenience.

cowtowncoder:
@djay-S My concern is whether this does what is being requested, and from @membersound 's comment I suspect it does not: removing column from `CsvSchema` is different from having column in schema but ignoring it for deserialization.
Tests just check that columns are removed from `CsvSchema` but not behavior wrt using such schema.
I assume behavior for removed columns will be "No such property" wrt deserialization.

@membersound you may be able to try this out by building `2.16.0-SNAPSHOT` locally from this PR, in case you think this might work.

Leaving this open for now but may close in future unless it is thought to actually solve the problem.

cowtowncoder:
I don't think this makes sense, closing.

PR #427, opened 2023-08-31
close IOContexts

pjfanning:
needed after https://github.com/FasterXML/jackson-core/pull/1064

If I have time, I can come back and try to add tests that check that buffer recyclers are being recycled.

### DIFF
--- a/toml/src/main/java/com/fasterxml/jackson/dataformat/toml/TomlGenerator.java
+++ b/toml/src/main/java/com/fasterxml/jackson/dataformat/toml/TomlGenerator.java
@@ -120,20 +120,23 @@ final class TomlGenerator extends GeneratorBase
 
     @Override
     public void close() throws IOException {
-        super.close();
-        _flushBuffer();
-        _outputTail = 0; // just to ensure we don't think there's anything buffered
+        if (!isClosed()) {
+            super.close();
+            _flushBuffer();
+            _outputTail = 0; // just to ensure we don't think there's anything buffered
 
-        if (_out != null) {
-            if (_ioContext.isResourceManaged() || isEnabled(StreamWriteFeature.AUTO_CLOSE_TARGET)) {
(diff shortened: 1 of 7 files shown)


toml/src/main/java/com/fasterxml/jackson/dataformat/toml/TomlGenerator.java
```diff
@@ -120,20 +120,23 @@ public Version version() {
...
+                if (_ioContext.isResourceManaged() || isEnabled(StreamWriteFeature.AUTO_CLOSE_TARGET)) {
+                    _out.close();
+                } else if (isEnabled(StreamWriteFeature.FLUSH_PASSED_TO_STREAM)) {
+                    // If we can't close it, we should at least flush
+                    _out.flush();
+                }
             }
+            _ioContext.close();
+            // Internal buffer(s) generator has can now be released as well
+            _releaseBuffers();
```
cowtowncoder, line 138: This should be moved before `IOContext.close()` (not a huge deal, but it will likely return buffers into recycler, and that should happen before releasing recycler into pool).
I can move after merge.

cowtowncoder (approved):

cowtowncoder:
Ok, so this does not depend on change in `jackson-core`, yet; but when that is merged, we'd probably want to sort of undo some of  these changes, right?

I'll go ahead and merge this for time being, regardless.

pjfanning:
> Ok, so this does not depend on change in `jackson-core`, yet; but when that is merged, we'd probably want to sort of undo some of these changes, right?
> 
> I'll go ahead and merge this for time being, regardless.

if https://github.com/FasterXML/jackson-core/pull/1064 is merged, there will need to be some changes in jackson-dataformats-text

cowtowncoder:
Yes, I merged #1064.

PR #441, opened 2023-11-20
test: use correct test input as stated by testname

tlahn:
Use test input that seems appropriate given test name and neighboring cases


cowtowncoder:
Thank you @tlahn, that makes sense. I'll see that CI passes and will merge if so.

Issue #445, opened 2023-12-04
`YAMLParser` throws unexpected `NullPointerException` in certain number parsing cases


arthurscchan:
In `YAMLParse::getNumberValueDeferred()` / `YAMLParse::_parseNumericValue()` / `YAMLParse::_parseIntValuev()` methods, the `lenght()` method of the String object `_cleanedTextValue` is called. This could cause an unexpected `NullPointerException` when the previous steps make `_cleanedTextValue` become null with an invalid input value. To fix it, it is suggested to add a conditional check and report an error when `_cleanedTextValue` is null.

```.java
if (_cleanedTextValue.length() <= 18) {
    return getNumberValue();
}
```

We found this issue by OSS-Fuzz and it is reported in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=64662.

PR #446, opened 2023-12-04
Fixes for issue #445: Wraps unexpected NullPointerException

arthurscchan:
This PR provides a possible fix for issue #445 by adding a conditional check to ensure `_cleanedTextValue` is not null even if the input in invalid.


yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLParser.java
```diff
@@ -1012,6 +1012,10 @@ public Object getNumberValueDeferred() throws IOException {
         // 01-Feb-2023, tatu: ParserBase implementation does not quite work
         //   due to refactoring. So let's try to cobble something together
 
+        if (_cleanedTextValue == null) {
+            _reportError("Current token ("+_currToken+") not numeric, can not use numeric value accessors");
```
cowtowncoder, line 1016: Huh? Based on what?

Let's start with basics: first a failing unit test, then see how to fix it. I think you have identified a problem, correctly, but proposed is probably masking the root cause.

cowtowncoder:
@arthurscchan In the meantime, one process thing: have we gotten a CLA from you? (I am not 100% sure; please LMK if you have).
If not, it'd be from here:

and is usually easiest to print, fill & sign, scan/photo, email to `cla` at fasterxml dot com.
This only needs to be done once before merging the first contribution; after that it's good for any and all contributions.

Looking forward to merging tests, fixes -- thank you for contributing these!

yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLParser.java
arthurscchan: https://github.com/FasterXML/jackson-dataformats-text/pull/446/commits/30414def0dab0fd2bae79d8c16f7501220801e09

cowtowncoder:
@arthurscchan Thank you for providing the test! I was able to change the resolution slightly, figuring out that test triggers buffering and thereby the reason for missing `_cleanedTextValue` -- problems is that code was not using previously decoded `int`/`long` value but assuming buffered number String.

PR #450, opened 2024-01-02
add package version

fivetran-abdulsalam:
Upgrade jackson version to 2.15


cowtowncoder:
I think title and description are wrong here: could you update @fivetran-abdulsalam ?

PR #465, opened 2024-02-27
YAML: consider starting `#` and ending `:` as quotable characters

MikeEdgar:
Scalar values starting with `#` or ending with `:` should be quoted. Currently, SnakeYAML will single-quote them automatically.  This change will proactively trigger the double quote style used for other values that require quotes when `MINIMIZE_QUOTES` is enabled.

### DIFF
--- a/release-notes/CREDITS-2.x
+++ b/release-notes/CREDITS-2.x
@@ -258,3 +258,8 @@ Mathieu Lavigne (@mathieu-lavigne)
 * Proposed #45 (and suggested implementation): (csv) Allow skipping ending line break
   (`CsvGenerator.Feature.WRITE_LINEFEED_AFTER_LAST_ROW`)
  (2.17.0)
+
+Michael Edgar (@MikeEdgar)
+
+* Contributed #465: (yaml) YAML: consider starting `#` and ending `:` as quotable characters
+ (2.17.0)
--- a/release-notes/VERSION-2.x
+++ b/release-notes/VERSION-2.x
@@ -14,6 +14,9 @@ Active Maintainers:
 === Releases ===
 ------------------------------------------------------------------------
 
+#465: (yaml) YAML: consider starting `#` and ending `:` as quotable characters
+ (contributed by Michael E)
+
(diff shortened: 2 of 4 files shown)


cowtowncoder (approved):
Looks good, will make some minor tweaks before merging but idea makes sense (change was subtle had to look carefully :) ).

cowtowncoder:
@MikeEdgar one procedural thing before merging: if we don't yet have CLA (only needs to be sent once for all contributions, before merging the first), it's here:

and it's usually easiest to print, fill & sign, scan/photo, email to `cla` at fasterxml dot com.
Once we have it I'll go ahead and merge this.

Thank you very much for contributing this (esp. great to have test coverage); looking forward to merging it.

oliversaggau:
I'm not quite sure if this is/was desired behavior, as the original description only states
> Scalar values starting with # or ending with : should be quoted.

But upgrading springdoc in our project to version 1.8.0 (which uses jackson version 2.17) in our project resulted in **all** scalar values being double-quoted in the generated .yaml file, not only the ones starting with # or ending with :

cowtowncoder:
@oliversaggau Thank you for double-checking :)

ivangal2:
@MikeEdgar , can you please explain/elaborate what was wrong with the single quotes before? And why do you think the double quotes are necessary for scalar values starting with # char?

MikeEdgar:
@ivangal2 it was inconsistent. Jackson would quote some values with `"` and SnakeYAML would detect other values and automatically use `'`. The point of this PR was to make use of one quoting style when quotes are automatically added.

Weinan-Zhang:
one question, what if single quote(') is needed to keep consistency, is there a way to configure the style of quote(single or double), thanks!

cowtowncoder:
@Weinan-Zhang  I am not 100% sure but I think there is no option for using single quote. If this is something you would like, feel free to file an issue requesting such feature (and if you have time and interest possibly even contribute a PR for adding). I don't know whether this would be easy to do -- considering complexity in configuring details of SnakeYAML output used by this module -- but it should be doable.

PR #470, opened 2024-04-01
Improve code quality by refactoring.

kavyabhojani:
Implementation smells:
1)Magic Number - CsvParser class
2)Complex condition - Parser class
3)Complex method - Parser class

Design Smells:
1)Multifaceted abstraction - CsvCharacterEscapes class


cowtowncoder:
I am not sure this is something we really want or need.

But it definitely could not go in a patch but would need to be against 2.18.

cowtowncoder:
Yeah I don't see the point of rewrite for sake of rewrite (for CSV parser); will close.

PR #471, opened 2024-04-02
Add a way to distinguish between null and empty

PujolDavid:
Following the issue: https://github.com/FasterXML/jackson-dataformats-text/issues/469

I made a pull request with that proposal:
- Add a new feature EMPTY_UNQUOTED_STRING_AS_NULL
- Implement it like EMPTY_STRING_AS_NULL
- Add tests

Note that I made this PR on the branch 2.18, but I'm using the version 2.15 in my projects (Spring boot 3.1)
If accepted we can move this feature also to previous versions

PR #472, opened 2024-04-04
Merge part of #471 ahead of it (to reduce diff)

cowtowncoder:


csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvParser.java
```diff
@@ -1448,4 +1433,23 @@ protected void _startArray(CsvSchema.Column column)
...
+     * 
+     * @since 2.17.1
+     */
+    protected boolean _isNullValue(String value) {
+        if (_nullValue != null) {
+            if (_nullValue.equals(value)) {
+                return true;
+            }
+        }
+        if (_cfgEmptyStringAsNull && _currentValue.isEmpty()) {
```
cowtowncoder, line 1450: _currentValue here is wrong: fixed in next commit.

PR #482, opened 2024-07-19
Allow passing `ParserImpl` by a subclass or overwrite the events

HeikoBoettger:
Makes it possible to overwrite the getEvent calls on the parser to add additional functionality such as filtering, modifying and inserting events. Alternatively provide a constructor to allow using subclasses of the snake yaml ParserImpl. Note: didn't replace the internal parserImpl-field since other subclasses might already use it otherwise would make sense to use the Parser-interface instead.

Why?

When using [networknt/json-schema-validator](https://github.com/networknt/json-schema-validator) with yaml files as data input format, I discovered that aliases and merges are not handled in a way that it is possible to validate the yaml files making use of this feature. While there are other discussions in the issue tracker to address this issue, I decided to manipulate the produced JsonTokens by instead of returning a VALUE_STRING producing JsonTolens as if the aliases were inlined the anchor. As it turned out it's simpler to hook into the getEvent-method and do the replacement on the snake yaml parser level. Since the code basically only invoked the getEvent-method of the parser in one place, I decided to extract a protected-method to overwrite it in my own code. This is the easiest way having to contribute a larger piece of code or having to completely copy the nextToken-method into my subclass.

Optionally I was thinking of injecting a subclass off the ParserImpl however that's also not possible, so I provide this option as well leaving other the choice how to extend the class.

### DIFF
--- a/yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLParser.java
+++ b/yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLParser.java
@@ -429,17 +429,6 @@ public class YAMLParser extends ParserBase
 
     // Note: SHOULD override 'getTokenLineNr', 'getTokenColumnNr', but those are final in 2.0
 
-    /**
-     * Since the parserImpl cannot be replaced allow subclasses to at least be able to
-     * influence the events being consumed.
-     *
-     * A particular use case is working around the lack of anchor and alias support to
-     * emit additional events.
-     */
-    protected Event getEvent() {
-        return _yamlParser.getEvent();
-    }
-    
     /*
     /**********************************************************
     /* Parsing
@@ -459,7 +448,7 @@ public class YAMLParser extends ParserBase
         while (true) {
             Event evt;
             try {
-                evt = _yamlParser.getEvent();
+                evt = getEvent();
             } catch (org.yaml.snakeyaml.error.YAMLException e) {
                 if (e instanceof org.yaml.snakeyaml.error.MarkedYAMLException) {
                     throw com.fasterxml.jackson.dataformat.yaml.snakeyaml.error.MarkedYAMLException.from
@@ -587,6 +576,19 @@ public class YAMLParser extends ParserBase
                 return _updateTokenToNull();
             }
         }
+    }
+
+    /**
+     * Since the parserImpl cannot be replaced allow subclasses to at least be able to
+     * influence the events being consumed.
+     *
+     * A particular use case is working around the lack of anchor and alias support to
+     * emit additional events.
+     *
+     * @since 2.18
+     */
+    protected Event getEvent() {
+        return _yamlParser.getEvent();
     }
 
     protected JsonToken _decodeScalar(ScalarEvent scalar) throws IOException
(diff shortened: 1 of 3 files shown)


yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLParser.java
```diff
@@ -190,18 +190,22 @@ public YAMLParser(IOContext ctxt, BufferRecycler br,
...
-
+    
     public YAMLParser(IOContext ctxt, int parserFeatures, int formatFeatures,
                       LoaderOptions loaderOptions, ObjectCodec codec, Reader reader)
+    {
+        this(ctxt, parserFeatures, formatFeatures, codec, 
+             new ParserImpl(new StreamReader(reader), (loaderOptions == null) ? new LoaderOptions() : loaderOptions))
+    }
+    
+    protected YAMLParser(IOContext ctxt, int parserFeatures, int formatFeatures,
```
cowtowncoder, line 201: Needs `@since 2.18` Javadoc; I can add.

yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLParser.java
```diff
@@ -415,12 +419,22 @@ protected JsonLocation _locationFor(Mark m)
...
     // Note: SHOULD override 'getTokenLineNr', 'getTokenColumnNr', but those are final in 2.0
 
+    /**
+     * Since the parserImpl cannot be replaced allow subclasses to at least be able to
+     * influence the events being consumed.
+     *
+     * A particular use case is working around the lack of anchor and alias support to
+     * emit additional events.
+     */
+    protected Event getEvent() {
```
cowtowncoder, line 429: This should probably be also called from somewhere? Most likely from

````
    public JsonToken nextToken() throws IOException  {

    }
````

will make change as well.

**Comment by cowtowncoder** [comment]
Hmmh. Code doesn't actually... compile. I'll fix it, but going forward it'd be good to at very least build the project before PR. :)

File: yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLParser.java [hunk 3]
```diff
@@ -434,7 +447,7 @@ public JsonToken nextToken() throws IOException
         while (true) {
             Event evt;
             try {
-                evt = _yamlParser.getEvent();
+                evt = getEvent();
```
cowtowncoder, line 450: This was missing: I think the idea was to allow overriding this way...

cowtowncoder:
@HeikoBoettger Made changes to get things to compile & the way I think you meant PR to be.

LMK if this is good; I understand the goal and adding overridability is fine. Can go in 2.18.0.

yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLParser.java
HeikoBoettger: Good catch.

HeikoBoettger:
@cowtowncoder Oh man, I just fired up the github web editor to get this quickly into a PR. I should have taken the time to clone the repository and checkin the actuall file I had locally. The idea with the constructor has a late idea.

Sorry for creating so much work on your side, that wasn´t my intention. Thanks for getting this ready.

cowtowncoder:
@HeikoBoettger no problem; figured it must be something like that. And since it's your first contribution, CI didn't automatically start (to show compilation errors).

Issue #497, opened 2024-09-21
UTF8Reader throw "Need to move partially decoded character; buffer not modifiable" when read only one chinese char


mrtaolili:
jackson-dataformat-csv  2.12.4

[...]

PR #498, opened 2024-09-21
Add failing test for #497

cowtowncoder:


cowtowncoder:
note: likely affects YAML, TOML variants of `UTF8Reader` as well...

PR #537, opened 2025-02-20
snakeyaml 2.4

pjfanning:
* has a bug fix for unicode surrogate chars
* https://bitbucket.org/snakeyaml/snakeyaml/wiki/Changes

### DIFF
--- a/yaml/pom.xml
+++ b/yaml/pom.xml
@@ -34,7 +34,7 @@
     <dependency>
       <groupId>org.yaml</groupId>
       <artifactId>snakeyaml</artifactId>
-      <version>2.3</version>
+      <version>2.4</version>
     </dependency>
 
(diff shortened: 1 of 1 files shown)


cowtowncoder:
Happy to merge the update, but TBH, this really should not fix the issue -- given that the issue is in `com.fasterxml.jackson.dataformat.yaml.UTF8Reader`. I suspect something else wrt buffering might be changing things so as not to trigger the problem.

But I am ok for changing test to be expected to pass -- if there is still an issue (which I think is the case), it'll re-surface eventually.

Will merge, update linked-to issue appropriately.

cowtowncoder (approved):

katepol:
when this will be released?

pjfanning:
@katepol the easiest thing is to upgrade the snakeyaml jar in your own build. It could be a few more weeks before the Jackson 2.19.0 release is made.

PR #541, opened 2025-02-26
add stream test to csv test case 497

pjfanning:
#497 issue only seems to happen with byte array input - work ok with InputStream input.

When byte array input to UTF8Reader, the _inputSource is always null and the _ioContext is null which is what is checked in `canModifyBuffer()`. Is there any way that this code can be adjusted to work for byte array input - or should we take the byte array and wrap as ByteArrayInputStream? - which would also fix the broken test.

The code that causes the issue.
https://github.com/FasterXML/jackson-dataformats-text/blob/9a38ea20f2733d790e9457418efbb92578b64ae6/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/impl/UTF8Reader.java#L406-L418

Needed for a fuzz test to work but also breaks the 497 test.

### DIFF
--- a/csv/src/test/java/com/fasterxml/jackson/dataformat/csv/deser/FuzzCSVReadTest.java
+++ b/csv/src/test/java/com/fasterxml/jackson/dataformat/csv/deser/FuzzCSVReadTest.java
@@ -1,5 +1,6 @@
 package com.fasterxml.jackson.dataformat.csv.deser;
 
+import java.io.ByteArrayInputStream;
 import java.io.IOException;
 
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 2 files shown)


pjfanning:
@cowtowncoder the new test passes but removing this code fixes the pre-existing broken test.

```
https://github.com/FasterXML/jackson-dataformats-text/blob/9a38ea20f2733d790e9457418efbb92578b64ae6/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/impl/UTF8Reader.java#L406-L418
```

cowtowncoder:
If I remember correctly, it's not really so much the case of `InputStream` actually working, but it not hitting AAOBE -- since there's empty input buffer extending beyond content end.
Or maybe put another way: it working by reading out of actual content buffer.

One way to test this would be to pass `byte[]` where content doesn't end at buffer end but couple of bytes before, having empty slots at end.

I'll merge this: I understand the bug itself but don't have time to fix it. Adding more tests may not help with that part significantly.

Issue #545, opened 2025-02-28
removing the fix for invalid input in FuzzCSVReadTest fixes the broken support for valid input in #497


pjfanning:
https://github.com/FasterXML/jackson-dataformats-text/blob/d43289df0be12aa0d8ee9cea97405cca337049bf/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/impl/UTF8Reader.java#L406-L418

invalid input that we might accidentally treat as valid if we remove the code
https://github.com/FasterXML/jackson-dataformats-text/blob/d43289df0be12aa0d8ee9cea97405cca337049bf/csv/src/test/java/com/fasterxml/jackson/dataformat/csv/deser/FuzzCSVReadTest.java

valid input that we will now properly support if we remove the code
https://github.com/FasterXML/jackson-dataformats-text/blob/d43289df0be12aa0d8ee9cea97405cca337049bf/csv/src/test/java/com/fasterxml/jackson/dataformat/csv/tofix/UnicodeCSVRead497Test.java#L23

Would it be possible to consider removing the code and trying later to come up with another fix for the fuzz case?
I think it is better to fix #497 and later come back to the fuzz case.

pjfanning:
@cowtowncoder #546 seems to indicate the removal of the code is fine - the fuzz case still behaves correctly, just the exception message changes a little

cowtowncoder:
Merged #546 with all the rework; I think we can close this.

PR #546, opened 2025-02-28
csv: remove old workaround in UTF8Reader

pjfanning:
* relates to #545 
* the fuzz test still behaves very similarly - just the exception message changes a little
* fixes the test for #497

### DIFF
--- a/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/impl/UTF8Reader.java
+++ b/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/impl/UTF8Reader.java
@@ -84,8 +84,8 @@ public final class UTF8Reader
         _inputPtr = ptr;
         _inputEnd = ptr+len;
         _autoClose = autoClose;
-        // Unmodifiable if there is no stream to actually read
-        // (ideally should pass explicitly)
+        // Unmodifiable if there is no stream to actually read from
+        // (ideally caller should pass explicitly)
         _inputBufferReadOnly = (in == null);
     }
 
--- a/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/impl/UTF8Reader.java
+++ b/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/impl/UTF8Reader.java
@@ -1,7 +1,6 @@
 package com.fasterxml.jackson.dataformat.csv.impl;
 
 import java.io.*;
-import java.util.Arrays;
 
 import com.fasterxml.jackson.core.io.IOContext;
 
@@ -34,7 +33,7 @@ public final class UTF8Reader
      *
      * @since 2.19
      */
-    private boolean _inputBufferReadOnly;
+    private final boolean _inputBufferReadOnly;
 
     /**
      * Pointer to the next available byte (if any), iff less than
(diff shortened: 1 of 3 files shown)


csv/src/main/java/com/fasterxml/jackson/dataformat/csv/impl/UTF8Reader.java
```diff
@@ -403,19 +403,6 @@ private boolean loadMore(int available) throws IOException
         // Bytes that need to be moved to the beginning of buffer?
         if (available > 0) {
             if (_inputPtr > 0) {
-                if (!canModifyBuffer()) {
```
cowtowncoder, line 406: Why do you feel this is no longer needed? This is to prevent reader from modifying buffer caller gave, in case moving would occur.

If anything, looks like it should avoid copying even if `_inputSource` was not null.

pjfanning: The test pass without this check. The existing tests and we gain an extra working test (the 497 test).

I don't know how to rewrite this because I don't have a broken test to work with. My main aim is not to have the 497 test (which is a valid test) fail. I'm not against revisiting this but since this change breaks no tests (other than minor woring diff in exception message on 1 test) and fixes one broken test, I'd prefer to remove this check for now. If this was merged, I would follow up and remove the same check in the TOML and YAML code bases because there is no proof that this check helps there either.

cowtowncoder: It'd be nice to have a test to verify underlying buffer is not modified when it shouldn't be -- which is what check does -- but even without that we should not make change that will now unexpectedly do that.

Specifically this is the case where caller passes `byte[]` containing UTF-8 content: expectation is that this buffer is not modified. And as such it should not be.

pjfanning: * ok, so I extended the tests to see if array input gets modified and it does
* is there anyway that the parser could be modified to clone the array when we hit this case and to work with the cloned array - to avoid modifying the input array?

pjfanning: I added back the code check for whether the buffer is modifiable - but instead of throwing an exception, I clone the buffer to avoid modifying the original. The cloning only happens in rare cases but I think it is worth the runtime hit to fix the broken test.

csv/src/main/java/com/fasterxml/jackson/dataformat/csv/impl/UTF8Reader.java
```diff
@@ -96,15 +112,8 @@ public UTF8Reader(IOContext ctxt, InputStream in, boolean autoClose)
...
         _inputEnd = 0;
         _autoClose = autoClose; 
-    }
-
-    /**
-     * Method that can be used to see if we can actually modify the
-     * underlying buffer. This is the case if we are managing the buffer,
-     * but not if it was just given to us.
-     */
-    protected final boolean canModifyBuffer() {
```
cowtowncoder, line 106: Although `protected`, not usable from outside, so can just remove.

csv/src/main/java/com/fasterxml/jackson/dataformat/csv/impl/UTF8Reader.java
cowtowncoder: I went ahead and used the idea, just rewrote slightly. Thank you for test to verify buffer not modified!

Ideally wouldn't need to move things around, will see what I can do there. But at least I can go ahead and merge this now.

cowtowncoder: Actually, can just remove copying action altogether.

csv/src/test/java/com/fasterxml/jackson/dataformat/csv/deser/FuzzCSVReadTest.java
```diff
@@ -6,16 +6,19 @@
...
 import com.fasterxml.jackson.dataformat.csv.CsvMapper;
+import com.fasterxml.jackson.dataformat.csv.ModuleTestBase;
 
 import static org.assertj.core.api.Assertions.fail;
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
 
 /**
  * Collection of OSS-Fuzz found issues for CSV format module.
  */
-public class FuzzCSVReadTest extends StreamingCSVReadTest
```
cowtowncoder, line 15: Not sure why base class was chosen that way; was wrong. :)

cowtowncoder (approved):

PR #551, opened 2025-03-21
Fix implimentation smells

Shadow2121:
Fixed following implementation smells:

- Magic Number (in `StringQuotingChecker.nameHasQuotableChar()`)
- Complex Method (in `CsvEncoder.endRow()`)
- Complex Conditional (in `YAMLGenerator.writeString()`)

Checklist:

- [x] Is the project building?
- [x] Are all existing test cases passing?

### DIFF
--- a/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/impl/CsvEncoder.java
+++ b/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/impl/CsvEncoder.java
@@ -553,21 +553,8 @@ public class CsvEncoder
     public void endRow() throws IOException
     {
         // First things first; any buffered?
-        if (_lastBuffered >= 0) {
-            final int last = _lastBuffered;
-            _lastBuffered = -1;
-            for (; _nextColumnToWrite <= last; ++_nextColumnToWrite) {
-                BufferedValue value = _buffered[_nextColumnToWrite];
-                if (value != null) {
-                    _buffered[_nextColumnToWrite] = null;
-                    value.write(this);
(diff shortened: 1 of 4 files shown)


csv/src/main/java/com/fasterxml/jackson/dataformat/csv/impl/CsvEncoder.java
```diff
@@ -587,6 +574,25 @@ public void endRow() throws IOException
         _outputTail += _cfgLineSeparatorLength;
     }
 
+    private void handleBufferedValues() throws IOException {
```
cowtowncoder, line 577: Single use. No point in extracting method.

yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/util/StringQuotingChecker.java
```diff
@@ -17,6 +17,7 @@ public abstract class StringQuotingChecker
     implements java.io.Serializable
 {
     private static final long serialVersionUID = 1L;
+    private static final int SPACE_CODE_POINT = 0x0020;
```
cowtowncoder, line 20: While use of magic constants is non ideal this is a single case of well-known ascii code.

cowtowncoder (commented):
I think hassle of getting CLA etc outweighs benefits here.

cowtowncoder:
@Shadow2121 While I appreciate your interest, I think there is little benefit of merging this pr (since I'd need to get CLA anyway), as per my notes.

PR #580, opened 2025-10-22
Check for missed columns even if given and expected column count are the same

### LINKED ISSUE #579: Incorrect detection of missed columns in header line if columns reordering is enabled
In addition to #285.
I'm expecting CsvMappingException when reading csv file 
```csv
name,agee
Roger,18
```

using code
```java
var mapper = new CsvMapper();
var schema = CsvSchema.builder()
  .setUseHeader(true).setReorderColumns(true)
  .with(CsvReadFeature.FAIL_ON_MISSING_HEADER_COLUMNS) //enabled by default
  .addColumn("name").addColumn("age")
  .build();

MappingIterator<Map<String, Object>> it = mapper
  .readerFor(Map.class)
  .with(schema)
  .readValues("""
    name,agee
    Roger,18""");
it.nextValue();
```

Hovewer, it doesn't throw it, even though the wanted 'age' column is missing.

lodik:
Fixes #579

### DIFF
--- a/csv/src/main/java/tools/jackson/dataformat/csv/CsvParser.java
+++ b/csv/src/main/java/tools/jackson/dataformat/csv/CsvParser.java
@@ -655,15 +655,17 @@ public class CsvParser
                 _reportCsvReadError("Empty header line: can not bind data");
             }
         }
-        // [dataformats-text#285]: Are we missing something?
-        int diff = schemaColumnCount - newColumnCount;
-        if ((diff > 0) && CsvReadFeature.FAIL_ON_MISSING_HEADER_COLUMNS.enabledIn(_formatFeatures)) {
+
+        if (CsvReadFeature.FAIL_ON_MISSING_HEADER_COLUMNS.enabledIn(_formatFeatures)) {
             Set<String> oldColumnNames = new LinkedHashSet<>();
             _schema.getColumnNames(oldColumnNames);
             oldColumnNames.removeAll(newSchema.getColumnNames());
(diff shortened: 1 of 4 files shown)


cowtowncoder:
Looks good, thank you for providing this, @lodik !

Just one thing before I can merge the fix in: we need CLA (from https://github.com/FasterXML/jackson/blob/main/contributor-agreement.pdf). It needs to be sent just once before merging the first PR and is valid for any future contributions.

The usual way is to print the doc, fill & sign, scan/photo, email to `cla` as fasterxml dot com.

Thank you again!

lodik:
cla: ✔

Btw, can it be backported to 2.x version? Should I open a separate PR, or will you handle it? Thanks in advance!

cowtowncoder:
@lodik If we had started with `2.x` it'd be straight-forward (if not trivial) to merge forward, but for reverse we need separate PR (against `2.x`; I can resolve merge conflicts).
Actually since there is no API change, could alternatively do PR against `2.21` to get into 2.21.1 patch (2.x is now for eventual 2.22.0).

These days we will only backport selected fixes into Jackson 2.x, fwtw -- this one is fine to add, if you need it.

One thing to note: in `2.x` there are `CsvParser.Feature`; 3.0 changed these to `CsvReadFeature`s. Otherwise changes are the same.

If you could provide a PR against `2.21` I'd be happy to review & merge it too!

cowtowncoder (approved):

PR #585, opened 2025-11-04
Fixes #10: missing `null` writes with `Object[]`

### LINKED ISSUE #10: Missing `null` writes when writing `Object[]` values
(moved from https://github.com/FasterXML/jackson-dataformat-csv/issues/116 by @georgewfraser)

When writing arrays to CSV, if there is a null, columns get written out of order. Reproduced in 2.7.1 with this test: https://gist.github.com/georgewfraser/164647fbaca61aadf9ee : 

```java

    @Test
    public void objectArray() throws IOException {
        CsvSchema schema = CsvSchema.builder()
                                    .addColumn("a", CsvSchema.ColumnType.NUMBER)
                                    .addColumn("b", CsvSchema.ColumnType.NUMBER)
                                    .setUseHeader(true)
                                    .build();
        ObjectWriter writer = Config.CSV.writer(schema);
        StringWriter out = new StringWriter();
        SequenceWriter sequence = writer.writeValues(out);

        sequence.write(new Object[]{ 1, 2 });
        sequence.write(new Object[]{ null, 2 });
        sequence.write(new Object[]{ null, null });
        sequence.write(new Object[]{ 1, null });

        assertEquals("\"a\",\"b\"\n" +
                     "1,2\n" +
                     ",2\n" +
                     ",\n" +
                     "1,\n", out.toString());
    }
```

### LINKED ISSUE #116: Error handling "null" String when Feature.MINIMIZE_QUOTES is active

cowtowncoder:
Claude-assisted attempt to fix #10 -- not sure if valid, solves tests but... need to review carefully.

Also fully fixes #116 (was marked as fixed but had failing test)

### DIFF
--- a/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvGenerator.java
+++ b/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvGenerator.java
@@ -238,6 +238,14 @@ public class CsvGenerator extends GeneratorBase
     protected boolean _skipValue;
 
     /**
+     * Flag set when a row has just been finished, used to distinguish between
+     * null values within a row vs null rows.
+     *
+     * @since 2.21
+     */
+    protected boolean _justFinishedRow = false;
+
+    /**
      * Separator to use during writing of (simple) array value, to be encoded as a
      * single column value, if any.
      *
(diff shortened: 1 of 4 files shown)


cowtowncoder:
Fixed in:

* 2.21.0
* 3.0.3

PR #587, opened 2025-11-08
Implement #154: ability to map empty Doc into empty Object (instead of empty stream)

cowtowncoder:
Interesting take by Claude to implement #154.


cowtowncoder:
Fixed in

* 2.21.0
* 3.0.3

PR #596, opened 2025-12-11
Port `YAMLAnchorReplayingFactory` from 2.x and improve it to handle nested anchors

indyteo:
Hi,

This PR aims to bring back and fix/improve the `YAMLAnchorReplayingFactory` implemented in 2.x by @HeikoBoettger-KarlStorz in #502, which was not included in 3.x as the title suggests (cf. https://github.com/FasterXML/jackson-dataformats-text/pull/502#issuecomment-2557939127).

Although porting this from 2.x to 3.x was not the initial goal of this PR! I originally ran into an issue when using the 2.x version because it wasn't able to handle nested anchors (cf. example at the end). Then I discovered Jackson 3.x was out, and decided to update (because maybe updating fixes issue?). However, I realized that that factory wasn't even in 3.x 😆 So I finally decided to do both at the same time. Nevertheless, I've split it into 2 distinct commits, in case we want to backport that fix to 2.x as well.

I've made sure the author of the original commit was a co-author of this one, to maintain credits for his work (thank him again for that, minor this little nested anchor scenario, it worked well!).
I've added a unit test for nested anchors.

Example of nested anchor YAML document:
```yaml
value: &valAnchor 'text'
obj1: &objAnchor
  string: *valAnchor
obj2: *objAnchor
```

Before the fix, the `'text'` node event replayed on line 3 by the `*valAnchor` alias event was not recorded as part of the `&objAnchor` anchor definition, and thus not replayed at line 4, which caused a parsing failure because of unexpected or illegal sequence of events replayed.
Now, events get recorded EVEN if they are part of the replayed events of an alias, when they are in an anchor definition. From my understanding, it should not cause any weird recursion issue, because anchors can only be replayed once they are completely finished to be registered.

Thank you

### DIFF
--- a/yaml/src/main/java/module-info.java
+++ b/yaml/src/main/java/module-info.java
@@ -11,8 +11,7 @@ module tools.jackson.dataformat.yaml
     exports tools.jackson.dataformat.yaml.util;
 
     provides tools.jackson.core.TokenStreamFactory with
-        tools.jackson.dataformat.yaml.YAMLFactory,
-        tools.jackson.dataformat.yaml.YAMLAnchorReplayingFactory;
+        tools.jackson.dataformat.yaml.YAMLFactory;
     provides tools.jackson.databind.ObjectMapper with
         tools.jackson.dataformat.yaml.YAMLMapper;
 }
--- a/yaml/src/main/java/module-info.java
+++ b/yaml/src/main/java/module-info.java
@@ -11,7 +11,8 @@ module tools.jackson.dataformat.yaml
     exports tools.jackson.dataformat.yaml.util;
 
     provides tools.jackson.core.TokenStreamFactory with
-        tools.jackson.dataformat.yaml.YAMLFactory;
+        tools.jackson.dataformat.yaml.YAMLFactory,
+        tools.jackson.dataformat.yaml.YAMLAnchorReplayingFactory;
     provides tools.jackson.databind.ObjectMapper with
         tools.jackson.dataformat.yaml.YAMLMapper;
 }
--- a/release-notes/CREDITS
+++ b/release-notes/CREDITS
@@ -14,3 +14,9 @@ Andrey Somov (asomov@github)
  (3.1.0)
 * Contributed #590: (yaml) Upgrade to the latest version of SnakeYAML Engine (3.0.1)
  (3.1.0)
+
+Théo SZANTO (@indyteo)
+
+* Contributed #596: (yaml) Port `YAMLAnchorReplayingFactory` from 2.x and improve it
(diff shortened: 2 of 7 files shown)


cowtowncoder:
First of all, big thank you for contributing this @indyteo !

Second: although we do have CLA for original contribution/contributor, I think we still need one for you.
It's at: 

https://github.com/FasterXML/jackson/blob/main/contributor-agreement.pdf

and only needs to be sent once, before the first contribution and is good for all future PRs.
The usual way is to print it, fill & sign, scan/photo, email to `cla` at fasterxml dot com.

Once we have it, I can final review and merge in this PR.

Looking forward to merging!

HeikoBoettger-KarlStorz:
@indyteo This is awesome. Nice to see that this feature is also something that others find useful.

indyteo:
Hi @cowtowncoder, I already sent you my CLA a few days ago (2025-12-11 12:04 CET). Is there any problem with it? I had no printer available at that time so I filled it digitally, through.

cowtowncoder:
@indyteo Sorry, I must not have checked email alias. Will do that now. Digital filling perfectly fine.

cowtowncoder:
@indyteo I got it just like you said. Thank you for pro-active sending!

yaml/src/main/java/module-info.java
```diff
@@ -11,7 +11,8 @@
     exports tools.jackson.dataformat.yaml.util;
 
     provides tools.jackson.core.TokenStreamFactory with
-        tools.jackson.dataformat.yaml.YAMLFactory;
+        tools.jackson.dataformat.yaml.YAMLFactory,
+        tools.jackson.dataformat.yaml.YAMLAnchorReplayingFactory;
```
cowtowncoder: I think we should not expose this via SPI but only standard factory.

cowtowncoder:
@indyteo Looks good, almost ready to merge. Just one thing; see my note on `module-info.java`; I don't think we should register alternate factory for SPI.

yaml/src/main/java/module-info.java
indyteo: Ok, to be fair, I don't know very well how modules work and what does this `provides` do. I just saw it was there and thought maybe it was needed for something haha.
Removed it :+1:

cowtowncoder: Yea no worries, not an unreasonable take. I am not 100% sure how usage even is, but seems like one primary factory per module is the way to go -- we can always add the alternative if need be.

cowtowncoder (approved):
LGTM

Issue #601, opened 2026-01-12
Reader should allow separating plain `nullValue` and quoted value `"nullValue"`


dmitry-workato:
Hey!

It seems there is an inconsistency in Jackson csv parser.

It has nice configuration to separate empty string and quoted empty strings:

```
String csv = "c1,c2,c3\n\"\",,\"\"";
var schema = CsvSchema.emptySchema();
var mapper = new CsvMapper();
var reader = mapper.readerForArrayOf(String.class)
                .with(schema)
                .with(CsvFactory.builder()
                        .disable(CsvParser.Feature.EMPTY_STRING_AS_NULL)
                        .enable(CsvParser.Feature.EMPTY_UNQUOTED_STRING_AS_NULL)
                        .build()
                )
                .with(CsvParser.Feature.WRAP_AS_ARRAY);
try (MappingIterator<String[]> csvIterator = reader.readValues(new ByteArrayInputStream(csv.getBytes()))) {
            csvIterator.next();
            var row = csvIterator.next();
            assertNotNull(row[0]);
            assertNull(row[1]);
            assertNotNull(row[2]);
        }
```

It also has nice configuration to specify the "nullValue":

However, we cannot find a way to read a CSV row that has configured `nullValue` as actual column value of "nullValue".
I think, the above test with this configuration should also work:

```
String csv = "c1,c2,c3\n\"nullValue\",nullValue,\"nullValue\"";
var schema = CsvSchema.emptySchema().withNullValue("nullValue");
```

It seems the deserialization of "nullValue" actually separates between string value of "nullValue" and nullValue.
It would really nice if the reader could also do that.

Am I missing something? Please advice!

dmitry-workato:
@PujolDavid could you please kindly take a look at this issue? Is there any simple "fix" for this?

Related PR that added the separation of empty string and quoted empty string:
https://github.com/FasterXML/jackson-dataformats-text/pull/471

cowtowncoder:
Almost certain no such distinction exists, or was planned. Compared to empty String, likely challenging to implement, unfortunately.

dmitry-workato:
Hey @cowtowncoder thanks a lot for the fast reply. I naively assumed that this patch might be enough:

```
FasterXML/jackson-dataformats-text - (3.x) > git diff
diff --git a/csv/src/main/java/tools/jackson/dataformat/csv/CsvParser.java b/csv/src/main/java/tools/jackson/dataformat/csv/CsvParser.java
[...]

PR #602, opened 2026-01-19
Implement #601: allow requiring quoting of "null value" (`CsvReadFeature.ONLY_UNQUOTED_NULL_VALUES_AS_NULL`)

cowtowncoder:


csv/src/main/java/tools/jackson/dataformat/csv/CsvParser.java
```diff
@@ -1235,12 +1241,17 @@ protected void _startArray(CsvSchema.Column column)
     protected boolean _isNullValue(String value) {
         if (_nullValue != null) {
             if (_nullValue.equals(value)) {
-                return true;
+                // [dataformats-text#601]: If NULL_VALUE_UNQUOTED_AS_NULL is enabled,
+                // only treat unquoted values as null
+                if (!_cfgOnlyUnquotedNullValuesAsNull || !_reader.isCurrentTokenQuoted()) {
+                    return true;
+                }
```
dmitry-workato:
```suggestion
                return !_cfgOnlyUnquotedNullValuesAsNull || !_reader.isCurrentTokenQuoted();
```

dmitry-workato (approved):

csv/src/main/java/tools/jackson/dataformat/csv/CsvParser.java
cowtowncoder: Would seem to not work correctly if `nullValue` set to empty String. But that depends I guess on precedence of "empty as null" settings. So I guess it's fine.

PR #608, opened 2026-02-13
Fix issue with UTF-8 surrogate pair decoding

pjfanning:
https://issues.oss-fuzz.com/issues/476184822

This may also affect other UTF8 Readers in other Jackson text formats.

### DIFF
--- a/yaml/src/main/java/tools/jackson/dataformat/yaml/UTF8Reader.java
+++ b/yaml/src/main/java/tools/jackson/dataformat/yaml/UTF8Reader.java
@@ -2,6 +2,7 @@ package tools.jackson.dataformat.yaml;
 
 import java.io.*;
 import java.lang.ref.SoftReference;
+import java.util.Objects;
 
 /**
  * Optimized Reader that reads UTF-8 encoded content from an input stream.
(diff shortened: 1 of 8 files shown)


yaml/src/main/java/tools/jackson/dataformat/yaml/UTF8Reader.java
```diff
@@ -174,6 +174,9 @@ public int read(final char[] cbuf, final int start, int len) throws IOException
 
         // Ok, first; do we have a surrogate from last round?
         if (_surrogate >= 0) {
+            if (outPtr >= len) { // no space to write pending surrogate
```
cowtowncoder: Isn't this illegal thing to do? `read()` method must return at least 1 `char` if asked for 1 or more, or return `-1` to indicate end-of-stream, or throw Exception. 0 is only ever allowed to be returned for reads with `len` of 0:

https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/io/Reader.html#read(char%5B%5D,int,int)

> If len is zero, then no characters are read and 0 is returned; otherwise, there is an attempt to read at least one character. If no character is available because the stream is at its end, the value -1 is returned; otherwise, at least one character is read and stored into cbuf.

cowtowncoder: And returning `0` is something that can easily cause infinite loops when caller tries to read more, getting nothing (but no -1 marker) etc.

pjfanning: copilot says this:
https://github.com/pjfanning/jackson-dataformats-text/pull/4/changes/27bc5c4b1891c450e525deb3f008c98a0ba72c80#r2806341466

and https://github.com/pjfanning/jackson-dataformats-text/pull/4#discussion_r2806341466

pjfanning: @cowtowncoder I've reread the code and I think this is safe. Essentially, the issue can only happen if the user requests zero bytes in the read and it seems fair to return zero as return value in that case.
What the code is doing is if there is a cached _surrogate value that it passes it back on the next read and the issue that OSS Fuzz seems to run into is that they ask for len=0 bytes.
I'd prefer to keep the change in this PR but I could change the PR to only return 0 value if the user asks for len=0.

cowtowncoder: I agree it works, but I would prefer explicit check if length is `<= 0` and return 0 directly; just for readability. It should be first thing to check, not specific to existence of buffered surrogate.

pjfanning: I changed the checks to handle invalid input values and to return 0 only if len==0

cowtowncoder (approved):
Looks good; will see if we can easily update other 2 UTF8Readers

yaml/src/main/java/tools/jackson/dataformat/yaml/UTF8Reader.java
cowtowncoder: Excellent. Thank you! I will try to copy changes to CSV and TOML formats too.

cowtowncoder:
Applied fix to csv & toml too; will merge.

PR #609, opened 2026-02-13
Catch and rethrow yaml engine internal exception on generation

pjfanning:
see https://oss-fuzz.com/testcase-detail/4866995549175808

### DIFF
--- a/yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLGenerator.java
+++ b/yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLGenerator.java
@@ -798,7 +798,12 @@ public class YAMLGenerator extends GeneratorBase
         _emit(new DocumentEndEvent(false));
     }
 
-    protected final void _emit(Event e) {
-        _emitter.emit(e);
+    protected final void _emit(Event e) throws JacksonException {
+        try {
+            _emitter.emit(e);
+        } catch (org.snakeyaml.engine.v2.exceptions.YamlEngineException ex) {
+            // Wrap SnakeYAML exception as Jackson exception
+            throw new JacksonYAMLWriteException(this, ex.getMessage(), ex);
+        }
     }
 }
--- a/yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLGenerator.java
+++ b/yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLGenerator.java
@@ -25,6 +25,7 @@ import org.snakeyaml.engine.v2.common.ScalarStyle;
 import org.snakeyaml.engine.v2.common.SpecVersion;
 import org.snakeyaml.engine.v2.emitter.Emitter;
 import org.snakeyaml.engine.v2.events.*;
+import org.snakeyaml.engine.v2.exceptions.YamlEngineException;
 import org.snakeyaml.engine.v2.nodes.Tag;
 
 public class YAMLGenerator extends GeneratorBase
@@ -801,7 +802,7 @@ public class YAMLGenerator extends GeneratorBase
     protected final void _emit(Event e) throws JacksonException {
         try {
             _emitter.emit(e);
-        } catch (org.snakeyaml.engine.v2.exceptions.YamlEngineException ex) {
+        } catch (YamlEngineException ex) {
             // Wrap SnakeYAML exception as Jackson exception
             throw new JacksonYAMLWriteException(this, ex.getMessage(), ex);
         }
--- a/yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLGenerator.java
+++ b/yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLGenerator.java
@@ -805,6 +805,9 @@ public class YAMLGenerator extends GeneratorBase
         } catch (YamlEngineException ex) {
             // Wrap SnakeYAML exception as Jackson exception
             throw new JacksonYAMLWriteException(this, ex.getMessage(), ex);
+        } catch (UncheckedIOException ex) {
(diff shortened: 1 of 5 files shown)


yaml/src/test/java/tools/jackson/dataformat/yaml/ser/GeneratorExceptionHandlingTest.java
```diff
@@ -0,0 +1,102 @@
...
+            gen.writeEndObject();
+            fail("Should have thrown an exception");
+        } catch (JacksonYAMLWriteException e) {
+            // Expected: our custom exception wrapping the underlying failure
+            assertNotNull(e.getMessage());
+            // Verify it has the cause
+            assertNotNull(e.getCause());
+        } catch (org.snakeyaml.engine.v2.exceptions.YamlEngineException e) {
+            fail("Should not leak SnakeYAML exception: " + e);
+        } catch (Exception e) {
```
cowtowncoder: This should catch `JacksonException`, not any `Exception`?

yaml/src/test/java/tools/jackson/dataformat/yaml/ser/GeneratorExceptionHandlingTest.java
```diff
@@ -0,0 +1,102 @@
...
+            // Other Jackson exceptions are acceptable (e.g., if caught at a different layer)
+            // but not SnakeYAML exceptions
+            assertFalse(e.getClass().getName().contains("snakeyaml"),
+                    "Should not leak SnakeYAML exception: " + e);
+        }
+    }
+
+    /**
+     * Basic sanity test that normal generation still works.
```
cowtowncoder: We already have plenty of regular write tests, this is not needed.

pjfanning: removed

yaml/src/test/java/tools/jackson/dataformat/yaml/ser/GeneratorExceptionHandlingTest.java
pjfanning: reworked the test

yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLGenerator.java
```diff
@@ -798,7 +799,15 @@ protected void _emitEndDocument() throws JacksonException {
...
-    protected final void _emit(Event e) {
-        _emitter.emit(e);
+    protected final void _emit(Event e) throws JacksonException {
+        try {
+            _emitter.emit(e);
+        } catch (YamlEngineException ex) {
+            // Wrap SnakeYAML exception as Jackson exception
+            throw new JacksonYAMLWriteException(this, ex.getMessage(), ex);
+        } catch (UncheckedIOException ex) {
+            // Wrap UncheckedIOException (e.g., from WriterWrapper) as Jackson exception
```
cowtowncoder: Let's unwrap `IOException`, first. Second, let's use existing `JacksonIOException` (from `jackson-core`)

cowtowncoder: Done.

cowtowncoder:
@pjfanning Made small change to use existing `JacksonIOException` which is used by `jackson-core` and YAML parser side to wrap underlying `IOException`s.

Now: we don't really have test to show how `YamlEngineException` (specifically, `EmitterException`) would be thrown. So I'm bit torn b/w merging this as speculative fix, vs postponing addition of new wrapper type, handling. I guess I can do partial merge now.

cowtowncoder:
Ok: merged `JacksonIOException` wrapping part, leaving the rest here.
Would be great to reproduce `YamlEngineException` throw case, its handling.

pjfanning:
The OSS Fuzz trace is

```
== Java Exception: org.snakeyaml.engine.v2.exceptions.EmitterException: expected NodeEvent, but got DocumentEnd
--
  | at org.snakeyaml.engine.v2.emitter.Emitter.expectNode(Emitter.java:466)
  | at org.snakeyaml.engine.v2.emitter.Emitter$ExpectDocumentRoot.expect(Emitter.java:448)
  | at org.snakeyaml.engine.v2.emitter.Emitter.emit(Emitter.java:241)
  | at tools.jackson.dataformat.yaml.YAMLGenerator._emit(YAMLGenerator.java:802)
  | at tools.jackson.dataformat.yaml.YAMLGenerator._emitEndDocument(YAMLGenerator.java:798)
  | at tools.jackson.dataformat.yaml.YAMLGenerator.close(YAMLGenerator.java:326)
```

So we see that it is possible that org.snakeyaml.engine.v2.emitter.Emitter.expectNode can throw org.snakeyaml.engine.v2.exceptions.EmitterException under certain circumstances.

I'm not sure how to make snakeyaml do this but we have some level of proof that is possible to trigger something like this.

cowtowncoder:
@pjfanning Ok. Given stack trace, maybe co-pilot could attempt a reproduction. FWTW, it sounds like empty YAML document, or mis-match with events. But I don't know how Fuzz case actually works; does it generate "YAML" content to round-trip (read-then-write) or something.

But it does sound like there is probably a path that could trigger this.

pjfanning:
@cowtowncoder maybe it's not as important to worry about issues generating YAML or other data formats. Parsing is a more attackable API - it's harder to manipulate a scenario where issues in generators can be abused

cowtowncoder:
Yeah I don't really consider RuntimeException leakage a real attack vector anyway in Java, more a usability/correctness/ergonomic thing.

cowtowncoder (approved):

PR #624, opened 2026-03-27
Fix YAML anchor replaying parser producing invalid sequence of events when using the merge operator inside an anchor

indyteo:
Hi,

Few months ago, I contributed a change (#596) to the `YAMLAnchorReplayingParser` to port it to 3.x and also to fix an issue regarding nested anchors handling. I tested the fix in various scenarios, but it seems that I forgot one edge case: When using the merge operator (`<<`) to reference an anchor (`*ref`), while being inside the definition of another anchor (`&anchor`), the `MappingStartEvent` of the referenced anchor (`ref`) gets recorded in the other anchor (`anchor`), thus resulting in an invalid sequence of events (2 mapping starts for 1 mapping end) when replaying that other anchor (`*anchor`). Depending on when the second anchor was replayed, it may have resulted either in an unknown alias (`anchor` definition being considered incomplete yet) or a random parsing exception because of a token mismatch (a mapping start right after another mapping start, instead of having a field name in-between).

<details>
<summary>Some YAML examples that were failing</summary>

```yaml
ref: &ref
  some: value
obj1: &anchor
  # first mapping start here
  <<: *ref # the bug: an extra mapping start event is recorded here,
           # making the algorithm recording the anchor believe it needs to
           # wait for an extra mapping end before the anchor is complete
  other: value
  # first (and only) mapping end here (but anchor is not considered complete yet!)
obj2: *anchor # fails here saying that 'anchor' is unknown (very misleading!!)
```

```yaml
ref: &ref
  some: value
wrapper:
  # first mapping start here
  obj1: &anchor
    # second mapping start here
    <<: *ref # bug: still extra mapping start recorded here
    other: value
    # first mapping end here
  # second mapping end here (anchor is now complete)
obj2: *anchor # parsing exception here saying that a field name
              # was expected but a mapping start was received
              # It indeed replayed 2 mapping starts back to back!
```
</details>

The fix was fairly simple to apply once the issue was properly identified. *Identifying the issue, on the other hand... anyway* 😄
I've added a way to consume the next event without recording it, so the merge operator can drop the mapping start without it being recorded in case it was inside an anchor definition.
I made a unit test covering this.

<details>
<summary>Side note about the other commit</summary>
I've noticed a lot of redundancy in the unit tests when asserting the current token location. I made an helper method to refactor:

```java
TokenStreamLocation loc = jp.currentTokenLocation();
assertEquals(lineNr, loc.getLineNr());
assertEquals(columnNr, loc.getColumnNr());
assertEquals(charOffset, loc.getCharOffset());
assertEquals(byteOffset, loc.getByteOffset());
```
into:

```java
assertLocation(jp, lineNr, columnNr, charOffset, byteOffset);
```
It is isolated in another commit for easier review.
</details>

I am submitting this PR against the `3.1` base branch instead of the main `3.x` branch because I think this is just a minor patch, 3.1.1. Feel free to correct me if it needs to be rebased against the `3.x` base branch for the next 3.2 release.

Thank you

### DIFF
--- a/release-notes/CREDITS
+++ b/release-notes/CREDITS
@@ -20,6 +20,9 @@ Théo SZANTO (@indyteo)
 * Contributed #596: (yaml) Port `YAMLAnchorReplayingFactory` from 2.x and improve it
   to handle nested anchors
  (3.1.0)
+* Fixed #624: (yaml) Fix YAML anchor replaying parser producing invalid sequence of events
+  when using the merge operator inside an anchor
+ (3.1.1)
 
 Dmitry Bufistov (@dmitry-workato)
 
--- a/release-notes/VERSION
+++ b/release-notes/VERSION
@@ -22,6 +22,9 @@ No changes since 3.1
 
 #615: (csv) Feature `CsvReadFeature.EMPTY_UNQUOTED_STRING_AS_NULL` ignored
  (reported by Henning L)
+#624: (yaml) Fix YAML anchor replaying parser producing invalid sequence of events
+  when using the merge operator inside an anchor
+ (fix by Théo S)
 
 3.1.0 (23-Feb-2026)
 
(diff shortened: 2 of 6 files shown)


cowtowncoder:
@indyteo If it's safe fix, 3.1 is a good choice -- it's an LTS branch where we do backport fixes without high risk.

cowtowncoder (approved):

cowtowncoder:
Thank you @indyteo !

PR #627, opened 2026-03-29
Fix #623: Add support for YAML 1.2 core schema

### LINKED ISSUE #623: Add support for YAML 1.2 non-finite (Infinite, -Infinite, NaN) numbers, Octal notation
The YAML 1.2 specification brings YAML into compliance with JSON as an official subset (via the [JSON schema](https://yaml.org/spec/1.2.2/#102-json-schema)).  It also fixes many of the ambiguities in YAML 1.1; e.g. the "Norway problem" (via the [core schema](https://yaml.org/spec/1.2.2/#103-core-schema)). 

I'd like to be able to use YAML 1.2 with Jackson. More specifically, I'd like to be able to validate documents using the YAML 1.2 core schema via [networknt json-schema-validator](https://github.com/networknt/json-schema-validator), which depends on Jackson for its YAML support, which in turn depends on [snakeyaml](https://github.com/snakeyaml/snakeyaml), which only supports YAML 1.1. Its sister (replacement?) project [snakeyaml-engine](https://github.com/snakeyaml/snakeyaml-engine) is a YAML 1.2 parser. It's not clear to me if snakeyaml-engine also fully supports YAML 1.1, and thus not clear if Jackson could replace snakeyaml with snakeyaml-engine or would need to support both.

mjcarman:
Add support for reading/writing [YAML 1.2 core schema](https://yaml.org/spec/1.2.2/#core-schema). Implements #623.

- Use CoreScalarResolver when CoreSchema is specified via LoadSettings.
- Map YAML string notation for infinity and non-a-number (`.inf` and `.nan`) to Java notation (`Infinity` and `NaN`).
- Serialize infinity and NaN values as `.inf` and `.nan`, respectively.
- Add handling for octal integers prefixed with `0o`.
- Added tests for deserializing from/serializing to YAML 1.2 core schema.

### DIFF
--- a/yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLFactory.java
+++ b/yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLFactory.java
@@ -49,6 +49,12 @@ public class YAMLFactory
     protected final SpecVersion _version;
 
     /**
+     * YAML schema for underlying parser to follow, if specified.
+     * @since 3.2
+     */
+    protected final YAMLSchema _schema;
+
+    /**
      * Helper object used to determine whether property names, String values
      * must be quoted or not.
      */
@@ -98,6 +104,7 @@ public class YAMLFactory
         // 26-Jul-2013, tatu: Seems like we should force output as 1.1 but
         //  that adds version declaration which looks ugly...
         _version = null;
+        _schema  = null;
         _quotingChecker = StringQuotingChecker.Default.instance();
         _loadSettings = null;
         _dumpSettings = null;
@@ -107,6 +114,7 @@ public class YAMLFactory
     {
         super(src);
         _version = src._version;
+        _schema  = src._schema;
         _quotingChecker = src._quotingChecker;
         _loadSettings = src._loadSettings;
         _dumpSettings = src._dumpSettings;
@@ -121,6 +129,7 @@ public class YAMLFactory
     {
         super(b);
         _version = b.yamlVersionToWrite();
+        _schema  = b.yamlSchema();
         _quotingChecker = b.stringQuotingChecker();
         _loadSettings = b.loadSettings();
         _dumpSettings = b.dumpSettings();
--- a/yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLFactory.java
+++ b/yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLFactory.java
@@ -4,7 +4,11 @@ import java.io.*;
 
 import org.snakeyaml.engine.v2.api.DumpSettings;
 import org.snakeyaml.engine.v2.api.LoadSettings;
+import org.snakeyaml.engine.v2.api.LoadSettingsBuilder;
 import org.snakeyaml.engine.v2.common.SpecVersion;
+import org.snakeyaml.engine.v2.schema.CoreSchema;
+import org.snakeyaml.engine.v2.schema.FailsafeSchema;
+import org.snakeyaml.engine.v2.schema.JsonSchema;
 
 import tools.jackson.core.*;
 import tools.jackson.core.base.TextualTSFactory;
@@ -131,7 +135,30 @@ public class YAMLFactory
         _version = b.yamlVersionToWrite();
         _schema  = b.yamlSchema();
         _quotingChecker = b.stringQuotingChecker();
-        _loadSettings = b.loadSettings();
+
+        if (b.loadSettings() == null) {
+            LoadSettingsBuilder builder = LoadSettings.builder();
+            if (_schema != null) {
+                switch (_schema) {
+                    case FAILSAFE:
+                        builder.setSchema(new FailsafeSchema());
+                        break;
+                    case JSON:
+                        builder.setSchema(new JsonSchema());
+                        break;
+                    case CORE:
+                        builder.setSchema(new CoreSchema());
+                        break;
+                    default:
+                        break;
+                }
+            }
+            _loadSettings = builder.build();
+        }
+        else {
+            _loadSettings = b.loadSettings();
+        }
+
         _dumpSettings = b.dumpSettings();
     }
 
--- a/yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLFactoryBuilder.java
+++ b/yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLFactoryBuilder.java
@@ -9,11 +9,7 @@ import tools.jackson.dataformat.yaml.util.StringQuotingChecker;
 
 import org.snakeyaml.engine.v2.api.DumpSettings;
 import org.snakeyaml.engine.v2.api.LoadSettings;
-import org.snakeyaml.engine.v2.api.LoadSettingsBuilder;
 import org.snakeyaml.engine.v2.common.SpecVersion;
-import org.snakeyaml.engine.v2.schema.CoreSchema;
-import org.snakeyaml.engine.v2.schema.JsonSchema;
-import org.snakeyaml.engine.v2.schema.FailsafeSchema;
 
 /**
  * {@link tools.jackson.core.TSFBuilder}
@@ -314,25 +310,6 @@ public class YAMLFactoryBuilder
      * @return the {@code SnakeYAML} configuration to use when parsing YAML
      */
     public LoadSettings loadSettings() {
-        if (_loadSettings == null) {
-            LoadSettingsBuilder builder = LoadSettings.builder();
-            if (_schema != null) {
-                switch (_schema) {
-                    case FAILSAFE:
-                        builder.setSchema(new FailsafeSchema());
-                        break;
-                    case JSON:
-                        builder.setSchema(new JsonSchema());
-                        break;
-                    case CORE:
-                        builder.setSchema(new CoreSchema());
-                        break;
-                    default:
-                        break;
-                }
-            }
-            return builder.build();
-        }
         return _loadSettings;
     }
 
--- a/yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLFactoryBuilder.java
+++ b/yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLFactoryBuilder.java
@@ -102,6 +102,7 @@ public class YAMLFactoryBuilder
     public YAMLFactoryBuilder(YAMLFactory base) {
         super(base);
         _version = base._version;
+        _schema  = base._schema;
         _quotingChecker = base._quotingChecker;
         _loadSettings = base._loadSettings;
         _dumpSettings = base._dumpSettings;
--- a/yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLFactoryBuilder.java
+++ b/yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLFactoryBuilder.java
@@ -330,7 +330,7 @@ public class YAMLFactoryBuilder
                         break;
                 }
             }
-            _loadSettings = builder.build();
+            return builder.build();
         }
         return _loadSettings;
     }
--- a/yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLFactoryBuilder.java
+++ b/yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLFactoryBuilder.java
@@ -9,7 +9,11 @@ import tools.jackson.dataformat.yaml.util.StringQuotingChecker;
 
 import org.snakeyaml.engine.v2.api.DumpSettings;
 import org.snakeyaml.engine.v2.api.LoadSettings;
+import org.snakeyaml.engine.v2.api.LoadSettingsBuilder;
 import org.snakeyaml.engine.v2.common.SpecVersion;
+import org.snakeyaml.engine.v2.schema.CoreSchema;
+import org.snakeyaml.engine.v2.schema.JsonSchema;
+import org.snakeyaml.engine.v2.schema.FailsafeSchema;
 
 /**
  * {@link tools.jackson.core.TSFBuilder}
@@ -40,6 +44,16 @@ public class YAMLFactoryBuilder
      * </p>
      */
     protected SpecVersion _version;
+
+    /**
+     * YAML schema for underlying parser to follow, if specified;
+     * left as {@code null} for backwards compatibility (which means
+     * whatever default settings {@code SnakeYAML} deems best).
+     * <p>
+     *     Ignored if you provide your own {@code LoadSettings}.
+     * </p>
+     */
+    protected YAMLSchema _schema;
 
     /**
      * Configuration for underlying parser to follow, if specified;
@@ -201,6 +215,20 @@ public class YAMLFactoryBuilder
     }
 
     /**
+     * Method for specifying the YAML schema to use during parsing; if
+     * {@code null} passed, will let {@code SnakeYAML} use its default settings.
+     *
+     * @param schema YAML schema to use for parsing, if not-null;
+     *    {@code null} for default handling
+     *
+     * @return This builder instance, to allow chaining
+     */
+    public YAMLFactoryBuilder yamlSchema(YAMLSchema schema) {
+        _schema = schema;
+        return this;
+    }
+
+    /**
      * Configuration for underlying parser to follow, if specified;
      * left as {@code null} for backwards compatibility (which means
      * whatever default settings {@code SnakeYAML} deems best).
@@ -257,6 +285,10 @@ public class YAMLFactoryBuilder
         return _version;
     }
 
+    public YAMLSchema yamlSchema() {
+        return _schema;
+    }
+
     public StringQuotingChecker stringQuotingChecker() {
         if (_quotingChecker != null) {
             return _quotingChecker;
@@ -277,6 +309,25 @@ public class YAMLFactoryBuilder
      * @return the {@code SnakeYAML} configuration to use when parsing YAML
      */
     public LoadSettings loadSettings() {
+        if (_loadSettings == null) {
+            LoadSettingsBuilder builder = LoadSettings.builder();
+            if (_schema != null) {
+                switch (_schema) {
+                    case FAILSAFE:
+                        builder.setSchema(new FailsafeSchema());
+                        break;
+                    case JSON:
+                        builder.setSchema(new JsonSchema());
+                        break;
+                    case CORE:
+                        builder.setSchema(new CoreSchema());
+                        break;
+                    default:
+                        break;
+                }
+            }
+            _loadSettings = builder.build();
+        }
         return _loadSettings;
     }
 
--- a/yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLFactoryBuilder.java
+++ b/yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLFactoryBuilder.java
@@ -52,6 +52,8 @@ public class YAMLFactoryBuilder
      * <p>
      *     Ignored if you provide your own {@code LoadSettings}.
      * </p>
+     *
+     * @since 3.2
      */
     protected YAMLSchema _schema;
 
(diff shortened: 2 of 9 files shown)


cowtowncoder:
First of all, thank you for contributing this PR -- from quick look it looks legit.

One process thing before I can merge it is we need a CLA (unless sent before), it's:

https://github.com/FasterXML/jackson/blob/main/CLA-jackson-2026.pdf

and usually easiest to print, fill & sign, scan/photo, email to `cla` at fasterxml dot com.
Once we get it, CLA is good for all future contributions.

Thank you once again, looking forward to reviewing and merging this pr!

yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLGenerator.java
```diff
@@ -617,15 +617,41 @@ public JsonGenerator writeNumber(BigInteger v) throws JacksonException
     public JsonGenerator writeNumber(double d) throws JacksonException
     {
         _verifyValueWrite("write number");
-        _writeScalar(String.valueOf(d), "double", STYLE_SCALAR);
+        if (Double.isNaN(d)) {
```
cowtowncoder: These changes should probably be gated behind new `YAMLWriteFeature` (defaulting to `false` to keep old behavior).

yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLParser.java
```diff
@@ -126,6 +127,12 @@ public class YAMLParser extends ParserBase
      */
     protected boolean _emittingSyntheticEmptyObject;    
 
+    /**
+     * Patterns for matching YAML's notation for Infinity and NaN values.
+     */
+    protected final Pattern _patternInf = Pattern.compile("([-+]?)\\.(?:inf|Inf|INF)");
```
cowtowncoder: Could and should be `static`.

mjcarman:
> One process thing before I can merge it is we need a CLA (unless sent before),

Oops, I didn't see anything about that when looking around for a contribution guide. Sent.

yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLParser.java
mjcarman: Added `static` modifier.

yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLGenerator.java
mjcarman: Added `YAMLWriteFeature.USE_YAML_NONFINITE_NOTATION` for backward-compatibility. (I didn't do this originally because I consider the current behavior to be a bug.)

yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLWriteFeature.java
```diff
@@ -140,6 +140,18 @@ public enum YAMLWriteFeature implements FormatFeature
...
+    /**
+     * Whether non-finite float values (infinities and not-a-number) should be
+     * emitted using YAML notation ({@code .inf}, {@code -.inf}, {@code .nan})
+     * or Java notation ({@code Infinity}, {@code -Infinity}, {@code NaN}).
+     * <p>
+     *     Default value is {@code false} for backwards compatibility.
+     * </p>
+     * <p>
+     * Feature is disabled by default.
+     */
```
cowtowncoder, line 153: Add `@since 3.2`

yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLWriteFeature.java
```diff
@@ -140,6 +140,18 @@ public enum YAMLWriteFeature implements FormatFeature
...
+    USE_YAML_NONFINITE_NOTATION(false),
```
cowtowncoder: @mjcarman Thinking about this a bit more, maybe default of `true` is ok -- it does seem like a bug.
So while one can argue about compatibility it seems reasonable to require changing config for old "wrong" behavior.

yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLWriteFeature.java
mjcarman: Added

yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLWriteFeature.java
mjcarman: I like this decision. Changed default to `true`.

cowtowncoder:
Looks good for the most part. One thing that may be missing is `YAMLParser` override of

    public abstract boolean isNaN();

which needs to be available to check for NaN/+Inf/-Inf cases.

And should have matching tests to ensure it correctly indicates such cases (some deserializers, like one for `BigDecimal`, require it to avoid failures and/or give useful error messages).

yaml/src/test/java/tools/jackson/dataformat/yaml/deser/YAML12CoreSchema623Test.java
```diff
@@ -0,0 +1,102 @@
...
+
+import tools.jackson.dataformat.yaml.ModuleTestBase;
+import tools.jackson.dataformat.yaml.YAMLFactory;
+import tools.jackson.dataformat.yaml.YAMLMapper;
+
+// Tests deserialization of YAML 1.2 core schema.
+public class YAML12CoreSchema623Test extends ModuleTestBase {
+    private final LoadSettings SETTINGS = LoadSettings.builder().setSchema(new CoreSchema()).build();
+    private final YAMLFactory  FACTORY  = YAMLFactory.builder().loadSettings(SETTINGS).build();
+    private final YAMLMapper   MAPPER   = YAMLMapper.builder(FACTORY).build();
```
cowtowncoder: Does this test not work with/cover "default" `YAMLMapper` (with default `YAMLFactory`)?

I think that is important use case; most users just use defaults.

Conversely if and when CoreSchema is needed, should have option in `YAMLFactory.builder()` to allow using that without having to manually construct one -- that is, have convenience method (`.useCoreSchema()` ?) -- or to refer to Snakeyaml-engine types?

mjcarman:
> One thing that may be missing is `YAMLParser` override of
> 
> ```
> public abstract boolean isNaN();
> ```

I don't think it's necessary to override the `isNaN()` method, but the way I'd added support for `.inf` and `.nan` into `_cleanYamlFloat()` was a bit ugly. I changed it to call `resetAsNan()` instead of remapping the strings to Java's notation. If you'd still like tests I'll write them, but I'd need some guidance on just what to test.

See commit 875f942d89ec.

cowtowncoder:
>  If you'd still like tests I'll write them, but I'd need some guidance on just what to test.

Right, this requires iterating over `JsonToken`s exposed by `YAMLParser`, which you can construct from `YAMLMapper`, given simple YAML doc.
Typical pattern is from, say, `tools.jackson.dataformat.yaml.deser.StreamingYAMLParseTest`:

```
    @Test
    public void testBasic() throws Exception
    {
        final String YAML =
 "string: 'text'\n"
// ....
;
        try (JsonParser p = MAPPER.createParser(YAML)) {
          assertToken(JsonToken.START_OBJECT, p.nextToken());
          assertToken(JsonToken.PROPERTY_NAME, p.nextToken());
          assertEquals("string", p.currentName());
          assertEquals("string", p.getString());
          assertToken(JsonToken.VALUE_STRING, p.nextToken());
          assertEquals("text", p.getString());

          // and so on
       }
```

yaml/src/test/java/tools/jackson/dataformat/yaml/deser/YAML12CoreSchema623Test.java
mjcarman: The serialization test would work with the defaults but this one depends on the `CoreSchema`. More specifically, it depends on the `CoreScalarResolver` for the schema. I'm okay with masking the use of snakeyaml-engine, but the method should probably be something like `useSchema(YAMLSchema.CORE)`, with the other options being `FAILSAFE` and `JSON`.

mjcarman:
> Right, this requires iterating over `JsonToken`s exposed by `YAMLParser` [...]
> Typical pattern is from, say, `tools.jackson.dataformat.yaml.deser.StreamingYAMLParseTest`:

Ah, that's simple enough. I already had a sample YAML file that I'd used for informal testing. Added `YAML12CoreSchemaParseTest` in commit 41395b8eb4cc.

yaml/src/test/java/tools/jackson/dataformat/yaml/deser/YAML12CoreSchema623Test.java
cowtowncoder: I am not suggesting preventing use of snakeyaml-engine abstractions but rather adding alternative convenience wrapper, if that makes sense? So easy one for common case but "manual" alternative for full control.

cowtowncoder:
Thanks! I hope to review this again tomorrow, get merged; we are pretty close.

yaml/src/test/java/tools/jackson/dataformat/yaml/deser/YAML12CoreSchema623Test.java
mjcarman: I had to mull this over for a while but I think I came up with a solution that matches the design philosophy you're using. I created a `YAMLSchema` enumeration and added a `yamlSchema()` method to `YAMLFactoryBuilder` to set it. If the user hasn't provided their own load settings but has specified a YAML schema, the builder will generate a `LoadSettings` object with the corresponding snakeyaml `Schema` object. I'm not crazy about doing that in the accessor, but there didn't seem to be another place to do it.

yaml/src/test/java/tools/jackson/dataformat/yaml/deser/YAML12CoreSchemaParseTest.java
```diff
@@ -0,0 +1,163 @@
...
+        - +.INF
+        - -.inf
+        - -.Inf
+        - -.INF
+        nans:
+        - .nan
+        - .NaN
+        - .NAN
+        strings:
+        - hello, world!
```
mjcarman, line 65: Add values for Infinity and NaN as plain scalars.

mjcarman: Added string values:
```
- Infinity
- NaN
- 'null'
- 'false'
- '0'
- '-1.0'
- '0o644'
- '0xFF'
```

yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLFactoryBuilder.java
```diff
@@ -41,6 +45,16 @@ public class YAMLFactoryBuilder
...
     protected SpecVersion _version;
 
+    /**
+     * YAML schema for underlying parser to follow, if specified;
+     * left as {@code null} for backwards compatibility (which means
+     * whatever default settings {@code SnakeYAML} deems best).
+     * <p>
+     *     Ignored if you provide your own {@code LoadSettings}.
+     * </p>
+     */
```
cowtowncoder, line 53: `@since 3.2`

yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLFactoryBuilder.java
```diff
@@ -200,6 +214,20 @@ public YAMLFactoryBuilder yamlVersionToWrite(SpecVersion v) {
...
     }
 
+    /**
+     * Method for specifying the YAML schema to use during parsing; if
+     * {@code null} passed, will let {@code SnakeYAML} use its default settings.
+     *
+     * @param schema YAML schema to use for parsing, if not-null;
+     *    {@code null} for default handling
+     *
+     * @return This builder instance, to allow chaining
```
cowtowncoder, line 223: `@since 3.2`

yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLFactoryBuilder.java
```diff
@@ -277,6 +309,25 @@ public StringQuotingChecker stringQuotingChecker() {
      * @return the {@code SnakeYAML} configuration to use when parsing YAML
      */
     public LoadSettings loadSettings() {
+        if (_loadSettings == null) {
+            LoadSettingsBuilder builder = LoadSettings.builder();
```
cowtowncoder: I think this pattern is fine, but let's just build and return but not overwrite `_loadSettings`.
Yes, this means construction would occur for each accessor call, but Jackson is pretty frugal wrt calling accessors.

This keeps info on explicit/non-explicit settings around

yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLFactoryBuilder.java
cowtowncoder, line 104: Missing copy for `_schema`?

yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLSchema.java
```diff
@@ -0,0 +1,23 @@
+package tools.jackson.dataformat.yaml;
+
+/**
+ * Enum of YAML 1.2 schemas that can be used for parsing YAML documents.
+ */
```
cowtowncoder, line 7: `@since 3.2`

yaml/src/test/java/tools/jackson/dataformat/yaml/deser/YAML12CoreSchema623Test.java
```diff
@@ -0,0 +1,100 @@
...
+import org.junit.jupiter.api.Test;
+
+import tools.jackson.dataformat.yaml.ModuleTestBase;
+import tools.jackson.dataformat.yaml.YAMLFactory;
+import tools.jackson.dataformat.yaml.YAMLMapper;
+import tools.jackson.dataformat.yaml.YAMLSchema;
+
+// Tests deserialization of YAML 1.2 core schema.
+public class YAML12CoreSchema623Test extends ModuleTestBase {
+    private final YAMLFactory FACTORY = YAMLFactory.builder().yamlSchema(YAMLSchema.CORE).build();
```
cowtowncoder, line 14: +1

yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLSchema.java
```diff
@@ -0,0 +1,23 @@
+package tools.jackson.dataformat.yaml;
+
+/**
+ * Enum of YAML 1.2 schemas that can be used for parsing YAML documents.
```
cowtowncoder, line 4: Should probably indicate which schema is used by default (if any?)

cowtowncoder (commented):
Very close -- big thank you for being very responsive & humoring me. :)

yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLFactoryBuilder.java
mjcarman: Argh. Added.

yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLFactoryBuilder.java
mjcarman: Argh. Added.

yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLFactoryBuilder.java
mjcarman: I changed the implementation to not store a LoadSettings object which was as a default, but it would look (almost) identical from an external perspective, as calling `YAMLFactoryBuilder.loadSettings()` wouldn't return `null`. If you want to preserve those semantics, I'd need to move the default construction into the `protected YAMLFactory(YAMLFactoryBuilder b)` constructor. Do you have a preference?

yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLFactoryBuilder.java
mjcarman: Refactored to copy the `_schema` value to YAMLFactory instances so that it can be restored from them when instantiating a builder from a factory.

yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLSchema.java
mjcarman: Man, you'd think I'd learn... ;) Added.

yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLSchema.java
mjcarman: If you don't specify a value you get whatever snakeyaml does by default. I don't say that here because this is just the definition for the enumeration. I did include this in the javadoc for `YAMLFactoryBuilder.yamlSchema()`:

```
 * Method for specifying the YAML schema to use during parsing; if
 * {@code null} passed, will let {@code SnakeYAML} use its default settings.
 *
 * @param schema YAML schema to use for parsing, if not-null;
 *    {@code null} for default handling
```

yaml/src/test/java/tools/jackson/dataformat/yaml/deser/YAML12CoreSchema623Test.java
mjcarman, line 1: @cowtowncoder Should I delete this and the matching test in 'ser'? This is pretty redundant with the 'deser' test you requested. Tests are nice, but tests that don't demonstrate anything useful just make the test suite slow.

yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLFactoryBuilder.java
mjcarman: I withdraw my question. Moving creation of a default LoadSettings object from the builder to the factory seems like a cleaner implementation, so I did it.

cf. commit 822544902405

yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLFactory.java
```diff
@@ -121,8 +133,32 @@ protected YAMLFactory(YAMLFactoryBuilder b)
     {
         super(b);
         _version = b.yamlVersionToWrite();
+        _schema  = b.yamlSchema();
         _quotingChecker = b.stringQuotingChecker();
-        _loadSettings = b.loadSettings();
+
+        if (b.loadSettings() == null) {
+            LoadSettingsBuilder builder = LoadSettings.builder();
```
cowtowncoder, line 140: What I meant was to keep construction in builder, but not assign local field, but leave it as `null`. But this works too.

yaml/src/test/java/tools/jackson/dataformat/yaml/deser/YAML12CoreSchema623Test.java
cowtowncoder: While redundant, I think it's ok. And this test in particular is quite fast.
Let me see the other one too, but at least keep this one

yaml/src/test/java/tools/jackson/dataformat/yaml/ser/YAML12CoreSchema623Test.java
```diff
@@ -0,0 +1,74 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import tools.jackson.dataformat.yaml.ModuleTestBase;
+import tools.jackson.dataformat.yaml.YAMLFactory;
+import tools.jackson.dataformat.yaml.YAMLMapper;
+import tools.jackson.dataformat.yaml.YAMLSchema;
+
+// Tests serialization of YAML 1.2 core schema.
+public class YAML12CoreSchema623Test extends ModuleTestBase {
```
cowtowncoder, line 17: Ok, `null` and `boolean` tests redundant, but we can keep it, not super big.

cowtowncoder (approved):

mjcarman:
@cowtowncoder Thank you for your time and support!

cowtowncoder:
Thank you @mjcarman for these improvements! Was great to work together to get them in.

Issue #669, opened 2026-05-17
TOML parser: invalid dotted-key extension of explicit tables / AOT not rejected


rajulbhatnagar:
## Summary
The toml-test invalid corpus has 5 fixtures testing the spec rule that an explicit `[table]` (or `[[array-of-tables]]`) cannot be extended afterwards via dotted keys at a higher scope. The current parser accepts them.

## Reproduction
Fixtures that should fail but currently succeed:
- `array/extend-defined-aot.toml`
- `table/append-with-dotted-keys-01.toml`
- `table/append-with-dotted-keys-02.toml`
- `table/append-with-dotted-keys-03.toml`
- `table/append-with-dotted-keys-08.toml`

## Spec citation
TOML 1.0.0 § "Defining a Table": "Defining tables out-of-order is discouraged. […] You don't need to specify all the super-tables if you don't want to (you can omit them). However, defining a [table] section with the same name as an already-defined table is not allowed."

PR #670, opened 2026-05-17
Implement #254: preserve TOML table syntax on round-trip

### LINKED ISSUE #254: More configurable TOML serialization
Right now, the toml serializer generates only top level properties, and inline tables where necessary (i.e. where arrays are used):

```toml
abc.foo = 1
abc.bar = 2
abc.xyz = [{foo = 1, bar = 2}]
```

There are special cases where we *could* generate normal toml tables. The example above can be expressed as:

```toml
[abc]
foo = 1
bar = 2

[[abc.xyz]]
foo = 1
bar = 2
```

This format will in many cases be more readable. However, there are problems with non-inline tables that prevent us from emitting them in a streaming generator:

- Once a table is started, it is impossible to go back to the parent or root table. If an object has scalar properties that we only see after we started a subtable, we have a problem.
- Array tables only work with arrays of objects. If an array is heterogeneous, but we've already started emitting it as an array table, we have a problem.

These problems can only be solved with knowledge of the tree being serialized, and perhaps even by influencing property order so that scalars always come first. However, because I expect serialization to toml to be fairly niche – it is a configuration format meant to be written by humans, after all – it is not worth adding machinery to databind to implement this.

Another question to consider: What even *is* the best representation of an object? A short inline object may be more readable than starting a new table that includes the entire path to that object.

A possible solution to this problem would be to add API to the TomlGenerator to specifically start a table. One approach would be to create overloads for `writeStartObject` and `writeStartArray` that allow forcing generation as a table. If data is then passed that cannot be represented as a table (the bullet points above), we would error. While we're at it, could also allow emitting comments with additional methods.

We could also inspect annotations on the `forValue` passed to `writeStartObject`, though this seems like misuse of that parameter, so not a good idea.

rajulbhatnagar:
## Summary

Resolves #254. Parsing a TOML document with `[section]` headers or
`[[array of tables]]` and writing the resulting tree back now preserves
those forms instead of collapsing to dotted keys.

## How it works

- **Origin-flag node subclasses + factory.** Package-private
  `TomlObjectNode` / `TomlArrayNode` carry `fromTableHeader` /
  `fromArrayOfTables` flags set by the parser. `TomlNodeFactory`
  produces them so `mapper.createObjectNode()` and `deepCopy()`
  round-trip the metadata. `TomlMapper._readTreeAndClose`
  short-circuits the deserializer reconstruction when the parser is
  one we created (a `TomlTreeTraversingParser` marker), preserving
  flags that the standard `JsonNodeDeserializer` path would have
  dropped. The short-circuit is disabled when the user has overridden
  the node factory, so a user-configured factory is honored (at the
  cost of flag preservation, the explicit opt-out trade-off).

- **Strict-ordering enforcement.** TOML rejects scalars after a
  sub-table at the same level. `TomlObjectNode.serialize` reorders
  children up front, deferring any branch that contains a table-
  flagged descendant. `TomlGenerator` also enforces the rule on the
  explicit-API path: scalars after `writeStartTable()` at any
  ancestor scope (not just the immediate context) throw
  `TomlStreamWriteException`.

- **Explicit generator API.** New public methods on `TomlGenerator`:
  `writeStartTable()`, `writeStartArrayOfTables()`. Callers building
  TOML from scratch can emit `[path]` and `[[path]]` directly without
  parsing first.

- **`BigDecimal` trailing-zero handling.** Honors
  `JsonNodeFeature.STRIP_TRAILING_BIGDECIMAL_ZEROES` on the
  short-circuit path so the override matches what
  `JsonNodeDeserializer` would have applied.

## Backward compatibility

Trees built via `mapper.createObjectNode()` with only scalar children
(no parsed `[section]` origin) still serialize as dotted keys. Verified
by `unflaggedNestedObjectStillUsesDottedKeys`. Existing users on
upgrade paths see no behavior change.

This branch is rebased on the current `3.x` HEAD, so it composes with
the recently-merged #669 fix (dotted key cannot extend an explicitly
defined table). The `explicitlyDefined` flag now lives on the promoted
`TomlObjectNode`; two tests
(`dottedKeyExtendingExplicitTableStillRejected`,
`explicitTableThenExplicitSubtableRoundTrips`) confirm the two features
coexist.

## Test plan

- `./mvnw -B -ntp -pl toml test` — all green (TomlRoundTripTest: 33
  cases; full module: 195 run, 2 skipped — the pre-existing `@Disabled`
  compliance stubs, untouched by this PR)
  - parsed -> write preservation, deepCopy preservation parameterized
    over `[table]`/`[[aot]]`, explicit-API emission, strict-ordering
    parameterized over 5 illegal-write cases, ancestor-scope scalar
    lockout, inline-scope fallthrough, custom-factory honor, BigDecimal
    feature on/off, `TreeTraversingParser` source-exposure contract,
    end-to-end fixture, #669 interaction

## Notes for reviewers

- The `TomlTreeTraversingParser` is a no-op marker subclass used only
  to discriminate parsers we created from caller-supplied
  `TreeTraversingParser` instances in `_readTreeAndClose`. Caller-
  supplied parsers go through the standard deserializer path so their
  source tree is not mutated (verified by
  `externalTreeTraversingParserDoesNotMutateSource`).
- The `_scalarsClosed` propagation walks ancestors up to the nearest
  enclosing `TABLE` / `ARRAY_OF_TABLES`, fixing a subtle issue where
  a `writeStartTable()` invoked from inside a `DOTTED_OBJECT` would
  not lock the root scope, allowing later root-level scalars to
  serialize inside the just-emitted section. Regression test:
  `scalarAtAncestorAfterNestedTableThrows`.
- The TOML 1.0.0 compliance test suite (toml-test) that surfaced
  several of the tracked parser bugs is intentionally **not** in this
  PR; it will be submitted separately.


yawkat:
Very intrusive changes for what is frankly a minor issue

cowtowncoder:
Definitely work needs to be split into smaller prs, 900+ file pr is a no-go.

rfscholte:
This topic has my interest. As stated in the description commit 1 is the compatibility test, which should be removed from this PR. (it makes more sense when there's a separate project that runs this testset again all toml parser, so it  can make comparison statistics)

cowtowncoder:
@rfscholte There's now #678 which I think adds compatibility testing suite (once fixes are in).

Or if not, separate PR for testing part would make sense.

cowtowncoder:
Compliancy test added for CI; compliancy deviations fixed.

This PR needs merging from `3.x` head; some conflicts to resolve.

rajulbhatnagar:
Ya makes sense for the compatibility suite to run as part of CI. Will remove that commit and rebase. Also happy to break this into more than 1 PR if you folks prefer

PR #674, opened 2026-05-18
Handle TOML byte order marks

### LINKED ISSUE #665: TOML parser: leading UTF-8 BOM rejected as 'Unknown token'
## Summary
TOML 1.0.0 § Spec allows (and many editors emit) a leading UTF-8 BOM (`U+FEFF`, bytes `EF BB BF`). The current parser rejects such files at byte 0 with `Unknown token`.

## Reproduction
toml-test fixtures `tests/valid/utf8-bom-01.toml`, `utf8-bom-02.toml`:
```
\xEF\xBB\xBF# This file starts with an UTF-8 BOM
key = "value"
```
```java
TomlMapper.builder().build().readTree(file);
// -> TomlStreamReadException: Unknown token at [Source: (File); line: 1, column: 1]
```

## Spec citation
TOML 1.0.0 spec: while the BOM is not part of the grammar proper, the toml-test corpus (the canonical compliance suite) lists these files under `valid/`. UTF-8 readers traditionally strip a leading BOM before lexing, which is what other TOML implementations do.

## Likely fix
In `TomlFactory._createParser(...Reader r)` (or the underlying UTF-8 reader path) skip a leading `U+FEFF` once before handing the stream to `Lexer`.

yawkat:
Handles TOML byte order marks according to the corpus expectations.

Changes:
- Accepts a single leading BOM at document start.
- Rejects BOM characters elsewhere in TOML input.
- Escapes BOM content when generating strings instead of emitting it literally.

Resolves #665

### DIFF
--- a/toml/src/main/java/tools/jackson/dataformat/toml/StringOutputUtil.java
+++ b/toml/src/main/java/tools/jackson/dataformat/toml/StringOutputUtil.java
@@ -62,6 +62,12 @@ class StringOutputUtil {
             return 0;
         }
 
+        // BOM is only permitted at the start of input. When writing it as content,
+        // force a basic string so it can be escaped instead of emitted literally.
+        if (c == 0xFEFF) {
+            return BASIC_STRING;
+        }
+
         // first, get the very restrictive unquoted keys out of the way.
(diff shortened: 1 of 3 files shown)


cowtowncoder (approved):
LGTM

PR #679, opened 2026-05-25
Validate integer length for hex/octal/binary radix literals

tonghuaroot:
The TOML `parseIntFromBuffer` decimal branch already routes the literal length through `StreamReadConstraints.validateIntegerLength` before falling back to `NumberInput.parseBigInteger`. The three radix-prefixed branches (`0x`, `0o`, `0b`) call `NumberInput.parseBigIntegerWithRadix` directly without that check, so configuring `maxNumberLength` does not constrain radix literals the way it does decimal ones.

This change applies `validateIntegerLength` once at the top of the radix branch, covering all three variants while leaving existing parsing logic unchanged. Error reporting reuses the same `[truncated]` shape as the decimal path.

Regression coverage added in `LongTokenTest` for each of the three prefixes against the default `TomlFactory` (max number length 1000).

2.x → 3.x backport left to maintainers.


cowtowncoder (approved):
LGTM

PR #680, opened 2026-05-25
Restore `validateIntegerLength()` check in `_decodeBigInt`

tonghuaroot:
On the 3.x branch, `YAMLParser._decodeBigInt` routes its input string straight to `NumberInput.parseBigInteger` / `parseBigIntegerWithRadix` without first running it through `StreamReadConstraints.validateIntegerLength`. The configured `maxNumberLength` cap therefore does not constrain integer literals that exit through this helper, including hex / octal / binary `!!int` radix scalars and any decimal scalar long enough to fall through to it.

This change adds the length check at the top of the method, mirroring the 2.x implementation and the surrounding number-validation helpers (`_cleanYamlInt`, `_cleanYamlFloat`) in the same file.

Regression coverage added in `IntegerLengthRadixTest` for decimal, hex, octal, and binary `!!int` literals against a `YAMLFactory` configured with `maxNumberLength(1000)`. Three of the four cases fail before the patch and pass after; the decimal one already passed via the existing `_cleanYamlInt` check and is included as a sanity guard.

2.x already has this check, so no backport is needed.

### DIFF
--- a/yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLParser.java
+++ b/yaml/src/main/java/tools/jackson/dataformat/yaml/YAMLParser.java
@@ -810,6 +810,7 @@ public class YAMLParser extends ParserBase
     }
 
     private BigInteger _decodeBigInt(String numStr, int base) throws JacksonException {
+        streamReadConstraints().validateIntegerLength(numStr.length());
         try {
             return base == 10 ?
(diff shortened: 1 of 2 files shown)


cowtowncoder (approved):
LGTM.

cowtowncoder:
Thank you @tonghuaroot !

PR #682, opened 2026-05-27
Clarify TOML array table comment

yawkat:
## Summary
- clarify the TOML parser comment for array-of-tables traversal
- remove the stale claim that simple dotted-key assignments into arrays of tables are accepted
- leave parser behavior unchanged

## Verification
- `JAVA_HOME=/usr/lib/jvm/java-21-openjdk MAVEN_USER_HOME=/home/opc/dev/agent-work/toml329/.m2-jackson ./mvnw -pl toml -am test -q`
- `git diff --check`


yawkat:
Merging without review since this is just a comment change

PR #697, opened 2026-07-01
Fix #696: write non-finite TOML floats as nan/inf/-inf

### LINKED ISSUE #696: `TomlMapper` writes `NaN`/`Infinity` as bare tokens, producing TOML it cannot read back
Was making some hardcore tests on my [EveryConfig](https://github.com/evernife/EveryConfig) and got a problem on the TOML parser.

## The problem

When a value is a non-finite `double` (NaN, positive infinity, or negative infinity), `TomlMapper` writes it using the Java text form `NaN`, `Infinity`, or `-Infinity`. That is not valid TOML. The TOML spec uses `nan`, `inf`, and `-inf` instead. Because of this, the output produced by the writer cannot be parsed again by the same `TomlMapper`: reading it back fails with `Unknown token`.

This is a writer-only problem. The reader already understands the correct TOML tokens `nan`, `inf`, `+inf`, and `-inf`, and reads them back as proper floating-point values. So writing `nan` / `inf` / `-inf` instead of the Java form would make these values round-trip, and it would need no change to the reader.

## How to reproduce

```java
import com.fasterxml.jackson.dataformat.toml.TomlMapper;
import java.util.Collections;

TomlMapper mapper = new TomlMapper();

// The writer produces a line that is not valid TOML.
String toml = mapper.writeValueAsString(Collections.singletonMap("x", Double.NaN));
System.out.println(toml);   // x = NaN

// Reading the writer's own output fails.
mapper.readTree(toml);      // throws TomlStreamReadException: Unknown token

// The reader does accept the spec form, which shows the fix is safe.
mapper.readTree("x = nan"); // ok: "x" is a number whose value is NaN
```

Reproduced with `jackson-dataformat-toml` 2.22.0 (latest at the time of writing). The behavior is the same on older 2.x releases.

Image

seonwooj0810:
Fixes #696

### Root cause
`TomlGenerator.writeNumber(double)` and `writeNumber(float)` render the value with `String.valueOf(...)`. For non-finite values that yields Java's textual forms `NaN`, `Infinity`, and `-Infinity`, which are **not** valid TOML floats. As a result the writer produces output it cannot read back — `TomlMapper.readTree(...)` on the writer's own output throws `TomlStreamReadException: Unknown token`.

### Change
Map non-finite values to the TOML float tokens `nan` / `inf` / `-inf` (a small shared helper), leaving finite values formatted exactly as before. This is a writer-only change: the parser already accepts `nan`, `inf`, `+inf`, and `-inf` (`Parser` line ~389), so written non-finite values now round-trip with no reader change.

### Tests
Added to `TomlGeneratorTest`:
- `nonFiniteDoubles` / `nonFiniteFloats` — assert exact output `abc = nan|inf|-inf`
- `nonFiniteRoundTrip` — writes NaN/±Infinity and reads them back through the same `TomlMapper`

All three fail before the change (two assertion failures plus a `Unknown token` error — the exact symptom in the issue) and pass after.

Verification done: `./mvnw -pl toml -am test -Dtest=TomlGeneratorTest` — 23 tests pass with the fix; reverting only the `TomlGenerator` change reproduces 2 failures + 1 `Unknown token` error.

### DIFF
--- a/toml/src/main/java/com/fasterxml/jackson/dataformat/toml/TomlGenerator.java
+++ b/toml/src/main/java/com/fasterxml/jackson/dataformat/toml/TomlGenerator.java
@@ -545,15 +545,35 @@ final class TomlGenerator extends GeneratorBase
     @Override
     public void writeNumber(double d) throws IOException {
         _verifyValueWrite("write number");
-        _writeRaw(String.valueOf(d));
+        _writeRaw(_nonFiniteTomlToken(d, String.valueOf(d)));
         writeValueEnd();
     }
 
     @Override
     public void writeNumber(float f) throws IOException {
         _verifyValueWrite("write number");
-        _writeRaw(String.valueOf(f));
+        _writeRaw(_nonFiniteTomlToken(f, String.valueOf(f)));
         writeValueEnd();
+    }
+
+    /**
+     * Maps a non-finite floating-point value to the TOML float token
+     * ({@code nan}, {@code inf} or {@code -inf}); finite values are written
+     * using the supplied Java text form. {@code String.valueOf(...)} would
+     * otherwise emit {@code NaN} / {@code Infinity} / {@code -Infinity}, which
+     * are not valid TOML and cannot be read back by the parser.
+     */
+    private static String _nonFiniteTomlToken(double d, String finiteForm) {
+        if (Double.isNaN(d)) {
+            return "nan";
+        }
+        if (d == Double.POSITIVE_INFINITY) {
+            return "inf";
+        }
+        if (d == Double.NEGATIVE_INFINITY) {
+            return "-inf";
+        }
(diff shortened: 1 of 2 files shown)


cowtowncoder:
Happy to merge if and when @yawkat approves.

yawkat:
Thanks!

cowtowncoder:
@yawkat @seonwooj0810 Going forward, we should update `release-notes/VERSION-2.x` (and maybe `CREDITS-2.x`) as part of pr.

yawkat:
Sorry, keep forgetting 😅

cowtowncoder:
@yawkat that happens no sweat. I can fix when I get home and can commit changes.

yawkat:
Added follow-up release-notes/credits PR: #698

PR #703, opened 2026-07-19
Add name-based column modification to `CsvSchema.Builder` (#699)

### LINKED ISSUE #699: Add ability to modify columns in `CsvSchema.Builder` by name
Currently you can only change a column by its index when building a schema, but there is no way to find any indices from the builder. You'd have to `build()` the current schema, find the index of the column you want to remove, then start a new builder that copies from it and hope the index hasn't changed.

It would make a lot more sense for the builder's fluent API to add `renameColumn(String, String)`, `removeColumn(String)`, etc.

seonwooj0810:
Fixes #699

`CsvSchema.Builder` only lets you modify columns by index (`renameColumn(int, ...)`, `replaceColumn(int, ...)`, `removeColumn(int)`), so modifying a column by *name* meant building the schema, scanning for the index, and rebuilding — and the index could shift in between. This adds name-based overloads so the fluent builder can locate a column by name directly.

## Changes
- `renameColumn(String oldName, String newName)`
- `replaceColumn(String name, Column c)`
- `removeColumn(String name)`

Each resolves the name to an index via a new `_columnIndex(String)` helper (linear scan, mirroring the existing `hasColumn(String)`) and delegates to the corresponding index-based method, so existing behavior is unchanged. An unknown name throws `IllegalArgumentException`, consistent with the existing `_checkIndex` out-of-range handling.

## Test evidence
Added to `CsvSchemaTest`:
- `testModifyColumnsByName` — rename/replace/remove by name preserve column positions and only touch the targeted column.
- `testModifyUnknownColumnByName` — an unknown name throws `IllegalArgumentException`.

`./mvnw -pl csv test -Dtest=CsvSchemaTest` → `Tests run: 11, Failures: 0, Errors: 0`.

Verification done: (1) no in-flight PR (searched open/all PRs for `CsvSchema renameColumn removeColumn`); (2) no self-claim (issue has no comments); (3) code-focused (`CsvSchema.java` + test); (4) confirmed the Builder had only index-based modifiers on `2.x`; (6) not a `spring-projects/*` repo, no triage gate.

### DIFF
--- a/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvSchema.java
+++ b/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvSchema.java
@@ -915,8 +915,11 @@ public class CsvSchema
         }
 
         /**
-         * Helper method for finding index of the column with given name; throws
-         * {@link IllegalArgumentException} if no column with the name exists.
+         * Helper method for finding index of the column with given name.
+         *
+         * @param name Name of the column to find
+         *
+         * @throws IllegalArgumentException if no column with given {@code name} exists
          *
          * @since 2.23
          */
--- a/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvSchema.java
+++ b/csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvSchema.java
@@ -630,6 +630,39 @@ public class CsvSchema
         }
 
         /**
+         * Method for renaming an existing column, located by its current name
+         * instead of by index. If no column with given {@code oldName} exists,
+         * an {@link IllegalArgumentException} is thrown.
+         *
+         * @since 2.23
+         */
+        public Builder renameColumn(String oldName, String newName) {
+            return renameColumn(_columnIndex(oldName), newName);
+        }
+
+        /**
+         * Method for replacing an existing column, located by its current name
+         * instead of by index. If no column with given {@code name} exists,
+         * an {@link IllegalArgumentException} is thrown.
+         *
+         * @since 2.23
+         */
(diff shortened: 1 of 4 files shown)


csv/src/main/java/com/fasterxml/jackson/dataformat/csv/CsvSchema.java
```diff
@@ -629,6 +629,39 @@ public Builder removeColumn(int index) {
             return this;
         }
 
+        /**
+         * Method for renaming an existing column, located by its current name
+         * instead of by index. If no column with given {@code oldName} exists,
+         * an {@link IllegalArgumentException} is thrown.
```
OrangeDog: Shouldn't it be
```
@throws IllegalArgumentException if no column with given {@code oldName} exists
```
?

seonwooj0810: Good catch — you're right. Converted the prose note into a proper `@throws IllegalArgumentException` tag (and added `@param` tags) on `renameColumn`, `replaceColumn` and `removeColumn` in 4037df12.

OrangeDog: (and `_columnIndex`)

seonwooj0810: Good point — done in 0440170d. Converted `_columnIndex`'s prose exception note into a proper `@throws IllegalArgumentException` tag and added a `@param name` tag, matching the public name-based methods.

cowtowncoder:
Added in 2.x (for 2.23.0) and 3.x (3.3.0).

PR #705, opened 2026-07-22
Fix #701: quote YAML 1.1 number-form Strings under ALWAYS_QUOTE_NUMBERS_AS_STRINGS

### LINKED ISSUE #701: `ALWAYS_QUOTE_NUMBERS_AS_STRINGS` does not quote YAML 1.1 exponent   (`1e5`), hex (`0x1F`) and underscore (`12_34`) number forms
## The problem

`YAMLGenerator.Feature.MINIMIZE_QUOTES` emits number-like Strings unquoted, producing a document that does not round-trip

## Env

**Module:** `jackson-dataformat-yaml` 2.22.0 (with `snakeyaml` 2.5)
**Java:** 25 (also reproduced on 8/11/17/21)
**Feature:** `YAMLGenerator.Feature.MINIMIZE_QUOTES`

Image

## Summary

With `MINIMIZE_QUOTES` enabled, a `String` value whose text looks like a YAML 1.1 number is emitted
**unquoted**. The resulting document no longer distinguishes that String from a number, which has two
observable consequences:

1. **Jackson's two read paths disagree on the same document.** Binding to a declared `String` returns the
   original text (the parser still has it), but reading through the untyped tree (`readTree` / `JsonNode`)
   returns the *resolved number*, so the text is gone.
2. **The document is not stable across a read+write cycle.** `writeValueAsString` → `readTree` →
   `writeValueAsString` turns `v: 1.10` into `v: 1.1`. The value changed, silently.

The write side is what is broken here: once `"1.10"` has been emitted as `v: 1.10`, the information that it
was a String is gone from the document, and *every* YAML consumer (Jackson's tree, SnakeYAML, PyYAML, Go's
yaml.v3) is correct to read it back as a float. The parser is not at fault; the emitter is.

That the intent to protect ambiguous scalars already exists is visible in two places: SnakeYAML refuses to
emit a leading-zero number plain (`'0176'`, the octal hazard), and Jackson's `StringQuotingChecker` quotes
the reserved keywords (`"true"`, `"null"`). Both stop short of the general case.

## The matrix (verified output of `printMatrix` below)

```
INPUT                    | EMITTED                  | TYPED READ   | TREE READ    | VERDICT
'176543210987654321'     | v: 176543210987654321    | 176543210987654321 | 176543210987654321 | ok
'9999999999999999999999' | v: 9999999999999999999999 | 9999999999999999999999 | 9999999999999999999999 | ok
'0176'                   | v: '0176'                | 0176         | 0176         | ok
'007'                    | v: '007'                 | 007          | 007          | ok
'00'                     | v: '00'                  | 00           | 00           | ok
'0'                      | v: 0                     | 0            | 0            | ok
'1.10'                   | v: 1.10                  | 1.10         | 1.1          | *** CORRUPTED ***
'1.0'                    | v: 1.0                   | 1.0          | 1.0          | ok
'3.14'                   | v: 3.14                  | 3.14         | 3.14         | ok
'1e5'                    | v: 1e5                   | 1e5          | 100000.0     | *** CORRUPTED ***
'0x1F'                   | v: 0x1F                  | 0x1F         | 31           | *** CORRUPTED ***
'12_34'                  | v: 12_34                 | 12_34        | 1234         | *** CORRUPTED ***
'+55'                    | v: +55                   | +55          | 55           | *** CORRUPTED ***
'-7'                     | v: -7                    | -7           | -7           | ok
'.5'                     | v: .5                    | .5           | 0.5          | *** CORRUPTED ***
'1:30'                   | v: 1:30                  | 1:30         | 1:30         | ok
'true'                   | v: "true"                | true         | true         | ok
'no'                     | v: "no"                  | no           | no           | ok
'null'                   | v: "null"                | null         | null         | ok
'~'                      | v: "~"                   | ~            | ~            | ok
'hello'                  | v: hello                 | hello        | hello        | ok
```

Note how many of the `ok` rows are **luck, not design**: `176543210987654321`, `0`, `1.0`, `3.14` and `-7`
are all emitted unquoted and re-read as numbers. They only survive because the canonical spelling of the
resolved number happens to equal the input text. Change `1.0` to `1.10` and the same code path corrupts it.

## `ALWAYS_QUOTE_NUMBERS_AS_STRINGS` fixes most of it, but not all

**It does not quote real numbers.** It only affects `writeString`, so genuine numeric values are untouched.
This test passes:

```java
realInt: 42
realLong: 176543210987654321
stringInt: "42"
plainWord: hello
```

**But its regex is too narrow.** `YAMLGenerator.PLAIN_NUMBER_P` is:

```java
protected final static Pattern PLAIN_NUMBER_P = Pattern.compile("[+-]?[0-9]*(\\.[0-9]*)?");
```

That matches `1.10`, `+55` and `.5`, which then get quoted and become safe. It does **not** match the
YAML 1.1 exponent, hex or underscore forms, so those stay unquoted and stay lossy even with the feature on:

```
v: 1e5     ->  tree read: 100000.0
v: 0x1F    ->  tree read: 31
v: 12_34   ->  tree read: 1234
```

So `ALWAYS_QUOTE_NUMBERS_AS_STRINGS` takes the failure count from 6/21 down to 3/21. It is a mitigation, not a fix, and it is opt-in, so the default configuration stays lossy.

## Reproducer

Single file, no custom modules, no annotations. Dependencies: `jackson-dataformat-yaml` + JUnit 5.

<details>

<summary> [[JAVA_CODE]] </summary>

```java
package com.example;

import com.fasterxml.jackson.core.type.TypeReference;
import com.fasterxml.jackson.dataformat.yaml.YAMLGenerator;
import com.fasterxml.jackson.dataformat.yaml.YAMLMapper;
import org.junit.jupiter.api.Test;

import java.util.Collections;
import java.util.Map;

import static org.junit.jupiter.api.Assertions.assertAll;
import static org.junit.jupiter.api.Assertions.assertEquals;

public class MinimizeQuotesTest {

    private static final YAMLMapper YAML = YAMLMapper.builder()
            .enable(YAMLGenerator.Feature.MINIMIZE_QUOTES)
            .disable(YAMLGenerator.Feature.WRITE_DOC_START_MARKER) // just to keep the assertions readable
            .build();

    /** String -> YAML -> tree -> String. Must be the identity for any input. */
    private static String roundTrip(String s) throws Exception {
        String yaml = YAML.writeValueAsString(Collections.singletonMap("v", s));
        return YAML.readTree(yaml).get("v").asText();
    }

    @Test
    public void ambiguousStringsAreEmittedUnquotedAndComeBackAsNumbers() throws Exception {
        assertEquals("v: 1.10", YAML.writeValueAsString(Collections.singletonMap("v", "1.10")).trim()); // ok: unquoted

        assertAll(
                () -> assertEquals("1.10", roundTrip("1.10")),        // FAILS -> "1.1"      canonical spelling differs
                () -> assertEquals("+55", roundTrip("+55")),        // FAILS -> "55"
                () -> assertEquals(".5", roundTrip(".5")),               // FAILS -> "0.5"
                () -> assertEquals("1e5", roundTrip("1e5")),         // FAILS -> "100000.0" YAML 1.1 exponent
                () -> assertEquals("0x1F", roundTrip("0x1F")),      // FAILS -> "31"       YAML 1.1 hex
                () -> assertEquals("12_34", roundTrip("12_34"))  // FAILS -> "1234"     YAML 1.1 underscore
        );
    }

    @Test
    public void theDocumentDoesNotSurviveOneReadWriteCycle() throws Exception {
        assertEquals("v: 1.10", YAML.writeValueAsString(YAML.readTree("v: 1.10")).trim()); // FAILS -> "v: 1.1"
    }

    @Test
    public void jacksonStillHasTheTextOnTheTypedReadPath() throws Exception {
        // Same document, other read path: proof the emitter, not the parser, threw the information away.
        Map<String, String> m = YAML.readValue("v: 1.10", new TypeReference<Map<String, String>>() {
        });
        assertEquals("1.10", m.get("v")); // ok -> declared String keeps the raw scalar text
    }
}
```
</details>

seonwooj0810:
Fixes #701

## Root cause
With `MINIMIZE_QUOTES` + `ALWAYS_QUOTE_NUMBERS_AS_STRINGS` enabled, `YAMLGenerator.writeString` decides whether a String must be quoted by matching it against `PLAIN_NUMBER_P` (`[+-]?[0-9]*(\.[0-9]*)?`). That regex only recognizes plain decimal integers/floats, so YAML 1.1 number forms it does **not** match — exponent (`1e5`), hex (`0x1F`), underscore (`12_34`) — were emitted **unquoted**. SnakeYAML (and every other YAML reader) then resolves those plain scalars back to numbers, so the value silently changes on a read/write cycle (`"1e5"` -> `100000.0`). The feature is documented to keep number-like Strings distinguishable, but its detection was incomplete.

## Change
Detection now also matches SnakeYAML's own implicit `int`/`float` resolver patterns (`Resolver.INT` / `Resolver.FLOAT`) — the exact patterns the parser uses to resolve plain scalars — in addition to the historical `PLAIN_NUMBER_P` (kept, so existing output is a strict superset and no previously-quoted value becomes unquoted). Any String that would resolve to a number is therefore quoted and round-trips. Scoped to the opt-in `ALWAYS_QUOTE_NUMBERS_AS_STRINGS` path; default `MINIMIZE_QUOTES` output is unchanged.

## Test evidence
Added `GeneratorWithMinimizeTest.testQuoteYAML11NumberFormsStoredAsString701`: asserts `1e5`, `0x1F`, `12_34`, `1.5e-3`, `0b101` are now quoted and round-trip through `readTree`, while a genuine non-number version String (`2.0.1.2.3`) stays unquoted. The test **fails before** this change (`1e5` emitted as `key: 1e5`) and **passes after**.

Verification done: full `yaml` module suite green (`./mvnw -pl yaml test` — 168 tests, 0 failures), including the pre-existing `ALWAYS_QUOTE_NUMBERS_AS_STRINGS` tests; the new test confirmed red-before/green-after.

### DIFF
--- a/release-notes/CREDITS-2.x
+++ b/release-notes/CREDITS-2.x
@@ -307,3 +307,15 @@ Sergio Delgado (@serandel)
 * Reported #154: (yaml) YAML file with no content throws `MismatchedInputException`
   when binding to Object type (POJO etc)
  (2.21.0)
+
+Pétrus Pradella (@EverNife)
+
+* Reported #701: (yaml) `ALWAYS_QUOTE_NUMBERS_AS_STRINGS` does not quote YAML 1.1
+  exponent (`1e5`), hex (`0x1F`) and underscore (`12_34`) number forms
+ (2.21.6)
+
+seonwoojung (@seonwooj0810)
+
+* Contributed fix for #701: (yaml) `ALWAYS_QUOTE_NUMBERS_AS_STRINGS` does not quote
+  YAML 1.1 exponent (`1e5`), hex (`0x1F`) and underscore (`12_34`) number forms
+ (2.21.6)
--- a/release-notes/VERSION-2.x
+++ b/release-notes/VERSION-2.x
@@ -16,6 +16,10 @@ Active Maintainers:
 
 2.21.6 (not yet released)
 
+#701: (yaml) `ALWAYS_QUOTE_NUMBERS_AS_STRINGS` does not quote YAML 1.1 exponent
+  (`1e5`), hex (`0x1F`) and underscore (`12_34`) number forms
+ (reported by @EverNife)
+ (fix contributed by @seonwooj0810)
 #702: (toml) Expand nesting depth checks for dotted keys
  (contributed by @yawkat)
 
--- a/yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLGenerator.java
+++ b/yaml/src/main/java/com/fasterxml/jackson/dataformat/yaml/YAMLGenerator.java
@@ -219,16 +219,6 @@ public class YAMLGenerator extends GeneratorBase
     protected final static long MIN_INT_AS_LONG = (long) Integer.MIN_VALUE;
     protected final static long MAX_INT_AS_LONG = (long) Integer.MAX_VALUE;
     protected final static Pattern PLAIN_NUMBER_P = Pattern.compile("[+-]?[0-9]*(\\.[0-9]*)?");
-
-    /**
-     * SnakeYAML's own implicit resolver patterns for {@code int} and {@code float}: used
-     * (in addition to {@link #PLAIN_NUMBER_P}) by {@link Feature#ALWAYS_QUOTE_NUMBERS_AS_STRINGS}
-     * to detect number-like Strings that {@link #PLAIN_NUMBER_P} misses -- notably YAML 1.1
-     * exponent ({@code 1e5}), hex ({@code 0x1F}) and underscore ({@code 12_34}) forms. Since these
-     * are the exact patterns the parser uses to resolve plain scalars back to numbers, quoting
-     * everything they match guarantees such Strings round-trip. See [dataformats-text#701].
-     */
-    protected final static Pattern[] NUMBER_RESOLVER_PS = { Resolver.INT, Resolver.FLOAT };
     protected final static String TAG_BINARY = Tag.BINARY.toString();
 
     /*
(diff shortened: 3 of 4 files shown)


cowtowncoder:
Could this be re-created against `2.21` branch?

seonwooj0810:
Done — rebased the fix onto `2.21` and re-targeted this PR's base branch to `2.21`. Cherry-picked cleanly (only the two files: `YAMLGenerator.java` +34/-1 and `GeneratorWithMinimizeTest.java` +26), and `GeneratorWithMinimizeTest` passes on 2.21 (14/14). Let me know if you'd prefer a brand-new PR instead of the re-target.

cowtowncoder (approved):

PR #713, opened 2026-08-13
Misc perf fixes (toml, yaml)

pjfanning:
suggested by Claude AI

## Performance Fixes

| File | Fix | Impact |
|------|-----|--------|
| `yaml/YAMLGenerator.java:691` | `String.valueOf(v.toString())` → `v.toString()` — eliminated redundant String conversion on BigInteger writes | High |
| `yaml/YAMLGenerator.java:327` | `Long.valueOf(id).toString()` → `Long.toString(id)` — avoids unnecessary boxing | Low |
| `yaml/YAMLGenerator.java:799` | Eliminated `Optional.of().map().Anchor()` chain → direct `new Anchor(id.toString())` | Medium |
| `toml/TomlParser.java:328-336` | Replaced `new String().replace().toCharArray()` (3 allocations) with single-pass in-place char copy skipping underscores | High |
| `toml/TomlParser.java:450` | Added `indexOf('_')` guard before `.replace("_", "")` — avoids unnecessary String copy on the common case (no underscores) | Medium |

### DIFF
--- a/toml/src/main/java/tools/jackson/dataformat/toml/TomlParser.java
+++ b/toml/src/main/java/tools/jackson/dataformat/toml/TomlParser.java
@@ -327,18 +327,17 @@ class TomlParser {
 
         for (int i = 0; i < length; i++) {
             if (buffer[start + i] == '_') {
-                // slow path to remove underscores: copy in-place, skipping '_'
-                char[] cleaned = new char[length];
-                int pos = 0;
-                for (int j = 0; j < length; j++) {
-                    char c = buffer[start + j];
+                // slow path to remove underscores: compact into the already-owned
+                // buffer itself (chars before the first '_' are already in place),
+                // avoiding a second array allocation
+                int pos = start + i;
+                for (int j = pos + 1; j < start + length; j++) {
+                    char c = buffer[j];
                     if (c != '_') {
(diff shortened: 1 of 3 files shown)


skrcode:
I ran JAIPilot Cloud against this exact PR head. It found one additional TOML allocation cleanup: compact underscores in place in the parser-owned token buffer instead of allocating a second char array.

The same 113 focused parser tests passed before and after, including a long token that forces buffer growth and a following token that checks for corruption. Five fixed allocation batches on a 10,000-digit underscored integer moved from 551,104 to 524,416 allocated bytes per parse, and the full 1,558-test clean build passed.

PR directly onto this source branch: https://github.com/pjfanning/jackson-dataformats-text/pull/13  
Cloud-generated draft and full evidence: https://github.com/skrcode/jackson-dataformats-text/pull/2

Only the underscore slow path changes; ordinary integer tokens are untouched, and the long-token fixture is not presented as a production-throughput claim.

pjfanning:
Thanks @skrcode - that change looks sensible and since it improves the change in my PR, it makes sense to merge it into this.

cowtowncoder:
Will now review, get merged.

cowtowncoder (approved):

PR #714, opened 2026-08-13
use varhandles for array access

pjfanning:
VarHandle array access removes JVM bounds checks on every array[index] operation. In a tight loop like the UTF8Reader ascii_loop that processes every byte of input, that's 2 eliminated bounds checks per iteration (1 read + 1 write). The JIT compiler can sometimes eliminate bounds checks on its own via loop unrolling and range analysis, but it depends on the loop shape — explicit VarHandle access guarantees the check is skipped regardless of how the JIT optimises the surrounding code.

The fallback ensures it still works on restricted JVMs (e.g. GraalVM native-image with --install-exit-handlers, or security managers that block MethodHandles.Lookup) — it just degrades to normal array access with bounds checks intact.

### DIFF
--- a/csv/src/main/java/tools/jackson/dataformat/csv/impl/BufferAccess.java
+++ b/csv/src/main/java/tools/jackson/dataformat/csv/impl/BufferAccess.java
@@ -1,0 +1,60 @@
+package tools.jackson.dataformat.csv.impl;
+
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.VarHandle;
+
+/**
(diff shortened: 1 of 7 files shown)


pjfanning:
the benefits here probably don't justify the extra code complexity

cowtowncoder:
Would definitely need basic JMH verification to see improvements -- not against changes, just want to know there are likely benefits.

pjfanning:
I may reopen but this but at the moment, I have other PRs that I want to concentrate on.

PR #718, opened 2026-09-08
Fix `AUTO_CLOSE_SOURCE` overrides and recycled-buffer leaks on read paths

pjfanning:
Four related resource-lifecycle defects found while auditing the read paths across all four modules. Each fix has a regression test that was verified to fail without it.

### 1. `AUTO_CLOSE_SOURCE` set on `ObjectReader` was ignored (properties, toml)

`JavaPropsFactory` and `TomlFactory` consulted the **factory's** `isEnabled(StreamReadFeature.AUTO_CLOSE_SOURCE)` instead of `readCtxt.getStreamReadFeatures(_streamReadFeatures)`. Enabling it on an `ObjectReader` left the caller's stream open; disabling it there still closed the caller's stream.

### 2. YAML left the caller's `InputStream` open (yaml)

`YAMLFactory._createReader()` baked the `UTF8Reader`'s `_autoClose` flag from the factory's feature set, but `YAMLParser._closeInput()` gates the close call on the *parser's*. When the two disagreed, the parser called `_reader.close()`, the Reader saw `_autoClose == false`, and the stream was silently kept open -- a real file-descriptor leak for file/socket sources:

```
factory default:                    parser AUTO_CLOSE=true, stream closed = true
factory disabled + reader enabled:  parser AUTO_CLOSE=true, stream closed = false   <-- leak
```

### 3. TOML lexer buffer leaked when the *first* token is invalid (toml)

`Lexer` takes a token buffer from the recycler in its constructor, but `TomlParser`'s constructor also lexed the first token -- outside the `try`/`finally` in `TomlParser.parse()` that calls `releaseBuffers()`:

```
input="" -> TomlStreamReadException
   at Lexer.yylex(Lexer.java:932)
   at TomlParser.<init>(TomlParser.java:42)
   token buffer: allocs=1, releases=0
```

Every other malformed input released correctly, so this was specifically the constructor window. The first `yylex()` now happens in `readFirstToken()`, called inside the `try`.

### 4. Factory-constructed `Reader`s not always closed (csv, properties, toml)

When the input is an `InputStream`, the factory wraps it in a `Reader` of ours (`UTF8Reader` / `Latin1Reader`) that holds a buffer from the recycler. That Reader was only closed when `AUTO_CLOSE_SOURCE` was enabled -- but these Readers already carry their own auto-close flag, so we can always close them and still leave the caller's stream alone. Previously the buffer was simply dropped whenever the read ended before EOF (`IOContext.close()` returns the `BufferRecycler` to the pool without sweeping buffers still held):

```
parse error partway through a 47 KB stream, AUTO_CLOSE_SOURCE off:
  user stream closed = false          (correct)
  read buffer: allocs=1, releases=0   <-- dropped
```

Reaching EOF frees the buffer on its own, which is why this only shows up on the abort path.

`Latin1Reader` gained the auto-close flag its csv/toml/yaml counterparts already had; it used to close the underlying stream unconditionally.

## API changes

All on `protected`/internal surface, and additive where a silent behaviour change was possible:

- `YAMLFactory._createReader(InputStream, JsonEncoding, IOContext)` now takes the effective stream-read features as a fourth argument. Changed rather than overloaded on purpose: an `@Override` of the old signature now fails to compile instead of being silently bypassed.
- `JavaPropsFactory._loadProperties(...)` likewise takes one extra argument.
- `CsvParser` and `CsvDecoder` gain constructors taking an `ownsReader` flag; the existing constructors delegate with `false`.

The UTF-16 branch of `CsvParserBootstrapper` is deliberately left as-is: it builds a plain `InputStreamReader`, which has no auto-close notion and would close the caller's stream regardless.

## Behaviour change worth flagging

`properties/StreamClosingTest` asserted the old behaviour of (1) -- `assertTrue(in.closed)` directly under a `// but not if reconfigured` comment. The comment describes the intent; the assertions now match it. Anyone relying on `.without(AUTO_CLOSE_SOURCE)` being a no-op for `JavaPropsMapper`/`TomlMapper` will see the stream stay open now.

## Tests

New: `toml/StreamClosingTest`, `toml/LexerBufferRecyclingTest`, `csv/deser/ParserBufferRecyclingTest`, `properties/ParserBufferRecyclingTest`, plus cases added to `yaml/deser/ParserAutoCloseTest`. The buffer tests assert against a `CountingRecyclerPool` test helper that counts alloc/release per buffer type, so a leak is a hard failure rather than a memory-pressure symptom.

Full suite green: csv 367, properties 74, yaml 269, toml 862.

### DIFF
--- a/csv/src/main/java/tools/jackson/dataformat/csv/impl/CsvDecoder.java
+++ b/csv/src/main/java/tools/jackson/dataformat/csv/impl/CsvDecoder.java
@@ -68,6 +68,17 @@ public class CsvDecoder
     protected boolean _autoCloseInput;
 
     /**
+     * Flag that indicates whether {@link #_inputReader} was constructed by Jackson
+     * (wrapping a caller-provided {@link java.io.InputStream}) rather than handed to
+     * us by the caller. Readers we construct must always be closed -- that is what
+     * returns their buffers to the recycler -- whereas the {@link java.io.InputStream}
+     * underneath is only closed if {@link #_autoCloseInput} is set.
+     *
+     * @since 3.3
+     */
+    protected final boolean _ownsInputReader;
+
+    /**
      * Configuration flag that determines whether spaces surrounding
      * separator characters are to be automatically trimmed or not.
      */
@@ -313,9 +324,23 @@ public class CsvDecoder
                       CsvSchema schema, TextBuffer textBuffer,
                       int stdFeatures, int csvFeatures)
     {
+        this(ctxt, owner, r, schema, textBuffer, stdFeatures, csvFeatures, false);
+    }
+
+    /**
+     * @param ownsReader Whether {@code r} was constructed by Jackson (and hence must
+     *    always be closed), or provided by the caller
+     *
+     * @since 3.3
+     */
+    public CsvDecoder(IOContext ctxt, CsvParser owner, Reader r,
+                      CsvSchema schema, TextBuffer textBuffer,
+                      int stdFeatures, int csvFeatures, boolean ownsReader)
+    {
         _owner = owner;
         _ioContext = ctxt;
         _inputReader = r;
+        _ownsInputReader = ownsReader;
         _textBuffer = textBuffer;
         _autoCloseInput = StreamReadFeature.AUTO_CLOSE_SOURCE.enabledIn(stdFeatures);
         _allowComments = CsvReadFeature.ALLOW_COMMENTS.enabledIn(csvFeatures);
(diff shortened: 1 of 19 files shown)


csv/src/main/java/tools/jackson/dataformat/csv/impl/CsvDecoder.java
```diff
@@ -312,10 +323,24 @@ public class CsvDecoder
     public CsvDecoder(IOContext ctxt, CsvParser owner, Reader r,
```
cowtowncoder, line 329: THis should probably be marked "@Deprecated" now?

properties/src/main/java/tools/jackson/dataformat/javaprop/io/Latin1Reader.java
cowtowncoder, line 68: Mark as deprecated.

cowtowncoder:
Good fixes but in future, please do separate PRs for diff backends when changes unrelated (easier for release notes).

csv/src/main/java/tools/jackson/dataformat/csv/impl/CsvDecoder.java
pjfanning: Done in 8cd0648b — marked `@Deprecated` (since 3.3) with a `@deprecated` Javadoc pointing at the replacement; no remaining callers.

properties/src/main/java/tools/jackson/dataformat/javaprop/io/Latin1Reader.java
pjfanning: Done in 8cd0648b — marked `@Deprecated` (since 3.3) with a `@deprecated` Javadoc pointing at the replacement; no remaining callers.

pjfanning:
Understood on splitting by backend — will do separate PRs per module for unrelated changes going forward (the later PRs from this batch, #721–#730, are already one module each). Both deprecation requests addressed in 8cd0648b.

cowtowncoder (approved):

PR #719, opened 2026-09-08
Fix output lost when writing to OutputStream with auto-close and flush both disabled

pjfanning:
Started as "the write-encoding buffer is never returned to the recycler when `AUTO_CLOSE_TARGET` is off". It turned out that buffer is not just leaked -- **the output in it is dropped**.

Properties and TOML have the same defect; originally fixed here too, now split out for per-module review: properties is #734, toml is #735. This PR covers csv only.

## The bug

When the write target is an `OutputStream`, `CsvFactory` wraps it in this module's own `UTF8Writer`, which buffers content. That wrapper was only closed when `AUTO_CLOSE_TARGET` was enabled, and only flushed when `FLUSH_PASSED_TO_STREAM` was enabled. With both disabled it was neither -- so whatever was still in its buffer never reached the stream. For any document smaller than that buffer, that is the entire document:

```
autoClose=true  flush=true  | JSON="{"a":"1","b":"2"}"  CSV="1,2\n"
autoClose=true  flush=false | JSON="{"a":"1","b":"2"}"  CSV="1,2\n"
autoClose=false flush=true  | JSON="{"a":"1","b":"2"}"  CSV="1,2\n"
autoClose=false flush=false | JSON="{"a":"1","b":"2"}"  CSV=""
```

`FLUSH_PASSED_TO_STREAM` is on by default, so reaching this needs an explicit `.disable(...)` of both -- but that is a legitimate combination (write several documents into one stream, flush once yourself at the end), and the failure is silent.

jackson-core's JSON generator is unaffected because it writes bytes straight to the `OutputStream` with no intermediate `Writer` -- its `_flushBuffer()` has already pushed everything out by the time `_closeInput()` runs. This module added a buffering layer and never drained it.

YAML happens to escape: SnakeYAML's emitter flushes on `StreamEndEvent`, so `YAMLGenerator` is covered by accident rather than design.

## The fix

A `Writer` we construct is ours to close, and closing it is what both flushes its content into the stream **and** returns its buffer to the recycler. So it is now always closed, with the caller's `OutputStream` protected separately. `UTF8Writer` is this module's own, so it gains an `autoClose` flag, exactly mirroring the one `UTF8Reader` in the same package already carries. `close()` always writes out pending bytes and releases the encoding buffer; it closes the `OutputStream` only if the flag is set.

`CsvEncoder.close()` now does:

```java
if (autoClose) {
    _out.close();
} else {
    if (flushStream) {
        _out.flush();
    }
    if (_ownsWriter) {
        _out.close();   // drains + recycles; the target is spared
    }
}
```

The write-encoding buffer balances again as a consequence:

```
before:  CSV AUTO_CLOSE_TARGET=false | writeEnc alloc=1 release=0
after:   CSV AUTO_CLOSE_TARGET=false | writeEnc alloc=1 release=1   (matching core JSON)
```

## API changes

`CsvGenerator` and `CsvEncoder` gain constructors taking an `ownsWriter` flag; the existing constructors delegate with `false`, so a caller-supplied `Writer` behaves exactly as before. `csv.impl.UTF8Writer` gains a three-arg constructor; the two-arg one delegates with `autoClose = true`.

## Not covered

`TokenStreamFactory._createWriter()` -- reached only via `createGenerator(out, enc)` with a non-UTF-8 encoding -- has the same defect, but the bug is in jackson-core, not here. Core JSON loses its output on that path too:

```
core JSON, UTF16_BE, autoClose=false flush=false -> bytes=0
core JSON, UTF8,     autoClose=false flush=false -> bytes=9
```

Worth a separate jackson-core issue.

## Tests

`GeneratorTargetClosingTest`, asserting across all four feature combinations that:

1. the content is fully written (the case that used to yield nothing),
2. the caller's stream is closed **iff** `AUTO_CLOSE_TARGET`,
3. draining our own writer does not turn into a `flush()` of the caller's stream when `FLUSH_PASSED_TO_STREAM` is off,
4. the stream *is* flushed when it is on.

Fails on `3.x` without the fix (on `autoClose=false, flushStream=false`). CSV suite green: 381 tests.

### DIFF
--- a/csv/src/main/java/tools/jackson/dataformat/csv/CsvFactory.java
+++ b/csv/src/main/java/tools/jackson/dataformat/csv/CsvFactory.java
@@ -303,11 +303,17 @@ public class CsvFactory
     protected CsvGenerator _createUTF8Generator(ObjectWriteContext writeCtxt,
             IOContext ioCtxt, OutputStream out)
     {
+        final int stdFeatures = writeCtxt.getStreamWriteFeatures(_streamWriteFeatures);
+        // Writer is constructed (and hence owned) by us, so the generator must always
+        // close it -- that flushes its pending content and recycles its buffer;
+        // `autoClose` only decides whether the caller's `OutputStream` goes with it
+        final boolean autoClose = ioCtxt.isResourceManaged()
+                || StreamWriteFeature.AUTO_CLOSE_TARGET.enabledIn(stdFeatures);
         return new CsvGenerator(writeCtxt, ioCtxt,
-                writeCtxt.getStreamWriteFeatures(_streamWriteFeatures),
+                stdFeatures,
                 writeCtxt.getFormatWriteFeatures(_formatWriteFeatures),
-                new UTF8Writer(ioCtxt, out), _getSchema(writeCtxt),
-                _characterEscapes, _maxQuoteCheckChars);
+                new UTF8Writer(ioCtxt, out, autoClose), _getSchema(writeCtxt),
+                _characterEscapes, _maxQuoteCheckChars, true);
     }
 
     private final CsvSchema _getSchema(ObjectWriteContext writeCtxt) {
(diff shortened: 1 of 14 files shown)


cowtowncoder:
@pjfanning  This PR needs to be split into multiple ones, per backend, before merging.

pjfanning:
I've cut this down to just CSV module and created new PRs for the other modules

cowtowncoder (approved):

PR #722, opened 2026-09-12
Use `NumberOutput` for number writes in Properties, TOML and YAML generators

pjfanning:
Follow-up to #721 (CSV), extending `NumberOutput` usage to the other three modules. Previously Properties, TOML and YAML all wrote numbers via `String.valueOf(...)` and ignored `StreamWriteFeature.USE_FAST_DOUBLE_WRITER` entirely.

**TOML** (`TomlGenerator`) and **Properties** (`WriterBackedGenerator`) own a `char[]` output buffer, so:
- `int`/`long` → `NumberOutput.outputInt/outputLong()` straight into the buffer
- `double`/`float` → `NumberOutput.outputDouble/outputFloat(char[])` (Schubfach) when `USE_FAST_DOUBLE_WRITER` is enabled; `NumberOutput.toString(v, false)` (= `Double.toString()`) otherwise
- TOML still maps non-finite values to `nan`/`inf`/`-inf` — the check now runs before any text is produced, instead of building a `String` first and discarding it

**Properties** base class `JavaPropsGenerator` gains non-abstract `_writeUnescapedEntry(int/long/double/float)` hooks with `String`-based defaults (`@since 3.3`), so `PropertiesBackedGenerator` — and any external sub-class — keeps working unchanged while `WriterBackedGenerator` overrides them.

**YAML** is bound to `String` scalars by the SnakeYAML emitter, so it now uses `NumberOutput.toString(v, useFast)` to honour the feature; non-finite handling (`.nan`/`.inf`) is unchanged.

Tests: new `TomlGeneratorNumberTest` and `NumberGenerationTest` (properties), plus fast-writer cases in `YAMLGeneratorNumberTest` — covering default vs. fast output, non-finite values, `Writer` and `OutputStream` targets, the `Properties`-backed generator, and runs that cross the output buffer boundary. Fast-path expectations are checked against `NumberOutput.toString(v, true)` since JDK < 19 `Float.toString()` isn't always shortest-repr.

### DIFF
--- a/properties/src/main/java/tools/jackson/dataformat/javaprop/JavaPropsGenerator.java
+++ b/properties/src/main/java/tools/jackson/dataformat/javaprop/JavaPropsGenerator.java
@@ -8,6 +8,7 @@ import java.util.Arrays;
 import tools.jackson.core.*;
 import tools.jackson.core.base.GeneratorBase;
 import tools.jackson.core.io.IOContext;
+import tools.jackson.core.io.NumberOutput;
 import tools.jackson.core.util.JacksonFeatureSet;
 
 import tools.jackson.dataformat.javaprop.io.JPropWriteContext;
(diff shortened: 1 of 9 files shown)


yawkat (approved):
only reviewed toml

cowtowncoder (approved):

PR #726, opened 2026-09-12
Fix TomlGenerator.writeString(char[],int,int) index bug when escaping needed

pjfanning:
`_writeStringImpl(int, char[], int, int)` reads `text[offset + len]` instead of `text[offset + i]` in the escaped basic-string branch. So `writeString(char[], int, int)` with any value containing a character that needs escaping (`"`, `\`, control chars) either writes the wrong characters (the one just past the range, repeated) or throws `ArrayIndexOutOfBoundsException` when the array has no slack after the range.

The `String` variant is unaffected — this only hits streaming users of the `char[]` overload, which is presumably why it went unnoticed (there were no tests for it).

Spotted while working on #725. One-line fix plus a regression test that checks the `char[]` output equals the `String` output and reads back correctly. Affects all branches from 2.18 through 3.x.


cowtowncoder:
Let's start winding down 2.18 branch and only backport security fixes. 2.21 may be targeted for correctness bugs like this one.

pjfanning:
retargeted to 2.21

PR #727, opened 2026-09-12
Avoid `char[]` copy of value text when parsing integer numbers in `CsvDecoder`

pjfanning:
Read-side counterpart to #721/#724.

`CsvDecoder.looksLikeInt()` began with `_textBuffer.contentsAsArray()`. Because the tokenizer has already materialized the column value as a `String` (`finishAndReturn()` → `contentsAsString()`), that call is `_resultString.toCharArray()` — a fresh copy of the token. `_parseIntValue()` then depended on that copy being exact-size (`int len = buf.length - offset`). Net effect: every numeric column accessed via `isExpectedNumberIntToken()` — which is how databind reads `int`/`long`/`BigInteger` properties from CSV — allocated a `char[]` before parsing.

Both methods now work on the `String` directly:
- `looksLikeInt()` scans via `charAt()`
- `_parseIntValue()` uses `NumberInput.parseInt(String)` (≤9 digits; JDK `Integer.parseInt` only for the rare `+` prefix, which `NumberInput` doesn't handle), `Long.parseLong` (10–18 digits, then the existing exactly-10-digit `int` narrowing), and `_parseSlowIntValue()` uses `NumberInput.inLongRange(String, neg)` for 19+ digits — the one remaining allocation is a `substring()` to strip the sign for that unsigned check, on 19+ digit signed values only
- the BigInteger path is unchanged (still lazy via `_numberString`)

The digit-count based selection of `int`/`long`/`BigInteger` is preserved exactly (e.g. `-000002147483648` is still reported as `LONG`, 22-char values in `long` range still go to `BigInteger`).

Side note: `contentsAsDouble()` already parsed from `_resultString`, so the float path had no copy — nothing to change there. TOML is already fully on `NumberInput` (including `parseLong19`/`inLongRange` on `char[]`); Properties doesn't parse numbers itself.

Tests: `CsvDecoderNumberTest` gains boundary cases for each branch — 9/10/18/19-digit values on both sides of the `int` and `long` limits, `+`/`-` signs, leading zeroes, `Long.MIN/MAX_VALUE ± 1` — plus a set of near-miss non-integers (`""`, `"-"`, `"1.0"`, `"1e5"`, `"0x10"`, `"--1"`…) that must stay `VALUE_STRING`. Full CSV suite passes.

### DIFF
--- a/csv/src/main/java/tools/jackson/dataformat/csv/impl/CsvDecoder.java
+++ b/csv/src/main/java/tools/jackson/dataformat/csv/impl/CsvDecoder.java
@@ -1315,15 +1315,17 @@ public class CsvDecoder
     }
 
     private boolean looksLikeInt() throws JacksonException {
-        final char[] ch = _textBuffer.contentsAsArray();
-        final int len = ch.length;
+        // NOTE: tokenizer has already materialized value as String (see
+        // `_nextUnquotedString()` etc), so access that: no copying needed
+        final String text = _textBuffer.contentsAsString();
+        final int len = text.length();
 
         if (len == 0) {
             return false;
         }
 
         int i = 0;
-        char c = ch[0];
+        char c = text.charAt(0);
         if (c == '-' || c == '+') {
             if (len == 1) {
                 return false;
@@ -1331,7 +1333,7 @@ public class CsvDecoder
             ++i;
         }
         for (; i < len; ++i) {
-            c = ch[i];
+            c = text.charAt(i);
             if (c > '9' || c < '0') {
                 return false;
             }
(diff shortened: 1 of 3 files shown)


pjfanning:
Before/after numbers (JMH 1.37, JDK 17.0.19, 3 forks × 5 warmup + 5 measurement iterations, `-prof gc`). Input: 1,000-row CSV with columns `int id, long ts, int qty, double price, String name`. "before" = `3.x` at b5b62c9b, "after" = this PR.

| Benchmark | | Time (µs/op) | Alloc (B/op) |
|---|---|---|---|
| `databindPojos` — POJO via `MappingIterator` | before | 562.8 ± 26.3 | 458,851 |
| | after | 555.9 ± 32.7 | 362,851 |
| | Δ | −1.2% (within noise) | **−21%** (−96 B/row) |
| `streamingInts` — `isExpectedNumberIntToken()` on every column | before | 423.2 ± 18.1 | 408,203 |
| | after | 400.0 ± 23.5 | 249,002 |
| | Δ | −5.5% (borderline) | **−39%** (−159 B/row) |

The allocation deltas match the mechanism exactly:
- databind: 3 integer columns per row → three `char[]` copies of ~4, 13 and ~3 chars = 24 + 48 + 24 = 96 B/row. (The `double` column already went through `NumberInput.parseDouble(String)`, so it never paid the copy.)
- streaming: `isExpectedNumberIntToken()` is called on all 5 columns, and previously `looksLikeInt()` copied the text *before* rejecting non-integers — so `price` and `name` paid ~32 B each as well: 96 + 64 ≈ 159 B/row.

Wall-clock differences are small and mostly inside the error bars (tokenizing + databind dominate; TLAB allocation of small arrays is cheap). The GC-pressure reduction is the real effect — GC count over the streaming run went 61 → 36.

<details>
<summary>Benchmark source</summary>

```java
@BenchmarkMode(Mode.AverageTime) @OutputTimeUnit(TimeUnit.MICROSECONDS)
@State(Scope.Benchmark) @Fork(3) @Warmup(iterations = 5, time = 1) @Measurement(iterations = 5, time = 1)
public class CsvReadBench {
    @JsonPropertyOrder({ "id", "ts", "qty", "price", "name" })
    public static class Row { public int id; public long ts; public int qty; public double price; public String name; }

    static final int ROWS = 1000;
    CsvMapper mapper; CsvSchema schema; ObjectReader reader; String csv;

    @Setup public void setup() {
        mapper = CsvMapper.builder().build();
        schema = mapper.schemaFor(Row.class).withHeader();
        reader = mapper.readerFor(Row.class).with(schema);
        StringBuilder sb = new StringBuilder("id,ts,qty,price,name\n");
        long ts = 1694530000000L;
        for (int i = 0; i < ROWS; i++) {
            sb.append(i).append(',').append(ts + i * 1000L).append(',').append((i * 7) % 500).append(',')
              .append((i % 100) + 0.25).append(',').append("item").append(i).append('\n');
        }
        csv = sb.toString();
    }

    @Benchmark public int databindPojos() throws Exception {
        int count = 0;
        try (MappingIterator<Row> it = reader.readValues(csv)) {
            while (it.hasNextValue()) { count += it.nextValue().qty; }
        }
        return count;
    }

    @Benchmark public long streamingInts(Blackhole bh) throws Exception {
        long sum = 0;
        try (JsonParser p = mapper.reader(schema).createParser(csv)) {
            JsonToken t;
            while ((t = p.nextToken()) != null) {
                if (t == JsonToken.VALUE_STRING) {
                    if (p.isExpectedNumberIntToken()) { sum += p.getLongValue(); } else { bh.consume(p.getString()); }
                }
            }
        }
        return sum;
    }
}
```
</details>

Issue #636, opened 2026-04-02
Support for `StreamReadConstraints.maxNameLength` and `maxStringLength` for YAML parsing


cowtowncoder:
(note: cleaved off of #430).

So: while some `StreamReadConstraints` constraints are enforced (like Max nesting depth, max number length), two constraints are not yet but should be supported:

1. Maximum name length
2. Maximum String value length

and it should be relatively easy to add these.

Issue #634, opened 2026-04-02
Support for `StreamReadConstraints.maxNameLength` for CSV parsing


cowtowncoder:
(note: cleaved off of #430).

So: while some `StreamReadConstraints` constraints are enforced (like Max nesting depth), and others irrelevant (max number length), three constraints should be supported:

1. Maximum name length (headers)
2. Maximum String value length (column values)
3. Maximum Document length

and it should be relatively easy to add these.

cowtowncoder:
Looks like (2) and (3) are already implemented, so just need to add (1).

PR #728, opened 2026-09-12
Support StreamReadConstraints maxNameLength/maxDocumentLength in TOML parser

### LINKED ISSUE #430: Support for StreamReadConstraints (max name, document lengths) -- TOML
https://github.com/FasterXML/jackson-core/blob/2.16/src/main/java/com/fasterxml/jackson/core/StreamReadConstraints.java 

equivalent of https://github.com/FasterXML/jackson-dataformat-xml/issues/609 and https://github.com/FasterXML/jackson-dataformats-binary/issues/392

In particular, the 2 new ones (in 2.16)

* maxDocumentLength
* maxNameLength

These settings are not likely to be supported for data formats in this project. Some support already exists for the pre-existing StreamReadConstraints settings.

pjfanning:
Fixes #430 — TOML was the last module here without these checks (CSV #634, YAML #636, Properties #638).

**`maxNameLength`** — validated in `TomlParser.parseAndEnterKey()` for each key part. Plain, quoted and dotted keys, table headers (`[a.b]`, `[[a]]`) and inline-table keys all flow through that method, so one check covers them all. String *values* are unaffected (they're covered by `maxStringLength` via the read-constrained `TextBuffer` already).

**`maxDocumentLength`** — the lexer keeps a running count of chars read from the input and calls `validateDocumentLength()` in `zzRefill()`. The hook is in `skeleton-toml` (two call sites: the bulk `read()` and the one-char low-surrogate read), with the counter and validation in the `toml.jflex` user-code section. Enforced regardless of input type (`String`/`Reader`/`InputStream`/`byte[]`), for limits smaller than the 4000-char lexer buffer, and for a single token spanning refills.

**Factory constraints for the tree parser** (second commit, folded in from #731) — TOML parses into an `ObjectNode` and exposes it via databind's `TreeTraversingParser`, which used to take its `StreamReadConstraints` from the `ObjectReadContext`. So the factory's constraints only applied when the caller was an `ObjectMapper`; with `tomlFactory.createParser(ObjectReadContext.empty(), …)` a configured `maxTokenCount` was silently ignored, unlike streaming parsers which always use their `IOContext`'s. `TomlFactory._createParser()` now passes `ctxt.streamReadConstraints()` explicitly, using the constructor added in FasterXML/jackson-databind#6216. With that, every `StreamReadConstraints` setting is enforced by the TOML backend regardless of entry point.

Tests: `TomlReadConstraintsTest` (11) — names at exactly the limit in every key position pass, one over fails in each position, values aren't affected; document limit within/over for all four input types, a limit below the buffer size, and a >10K single-token document. `TomlTokenCountTest` (4) — constraints identity, `maxTokenCount` via `createParser(ObjectReadContext.empty(), …)` for all five input types and via the mapper (2 of 4 fail without the fix). Full TOML suite passes (869).

Requires a jackson-databind `3.3.0-SNAPSHOT` containing FasterXML/jackson-databind@02e30645 (#6216, merged).

### DIFF
--- a/release-notes/VERSION
+++ b/release-notes/VERSION
@@ -16,6 +16,8 @@ implementations)
 
 3.3.0 (not yet released)
 
+#430: (toml) Support `StreamReadConstraints.maxNameLength()` and
+  `StreamReadConstraints.maxDocumentLength()` for TOML parser
 #699: (csv) Ability to modify columns in `CsvSchema.Builder` by name
  (requested by @OrangeDog)
  (fix by @seonwooj0810)
--- a/toml/src/main/java/tools/jackson/dataformat/toml/TomlParser.java
+++ b/toml/src/main/java/tools/jackson/dataformat/toml/TomlParser.java
@@ -153,6 +153,8 @@ class TomlParser {
             } else {
                 throw errorContext.atPosition(lexer).unexpectedToken(partToken, "quoted or unquoted key");
             }
+            // [dataformats-text#430]: each key part becomes a property name
+            tomlFactory.streamReadConstraints().validateNameLength(part.length());
             pollExpected(partToken, Lexer.EXPECT_INLINE_KEY);
             if (peek() != TomlToken.DOT_SEP) {
                 return new FieldRef(node, part, nodeDepth);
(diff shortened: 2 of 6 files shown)


yawkat (approved):

cowtowncoder:
Will re-create PR against 2.18.

cowtowncoder:
Re-created as #736.

PR #729, opened 2026-09-12
Support StreamReadConstraints.maxDocumentLength() in YAML parser

pjfanning:
Completes #636 (which delivered `maxNameLength`; `maxDocumentLength` was left out because SnakeYAML consumes the input). Same approach as TOML in #728: count at the input boundary.

`YAMLParser` hands SnakeYAML Engine a `Reader` (`new StreamReader(loadSettings, reader)`). When `StreamReadConstraints.hasMaxDocumentLength()`, that `Reader` is now wrapped in a small `ReadConstrainedReader` (`yaml.util`) that accumulates characters read/skipped and calls `validateDocumentLength()` after each read. When no limit is configured (the default) nothing is wrapped, so there's zero overhead; with a limit it's one `long` add + compare per buffer fill. Since every `_createParser` variant ends up passing a `Reader` (byte input goes through `UTF8Reader`), this covers `String`, `Reader`, `InputStream`, `byte[]` and `char[]` input in one place.

Note the limit applies to the whole stream, including multi-document YAML streams — same as for JSON's `MappingIterator` over a stream.

Tests: `YAMLDocumentLengthTest` — within/over the limit for all input types plus direct streaming access; a limit below SnakeYAML's read buffer size; a single scalar longer than the limit (fails on document length, not string length); multi-document streams; and explicit no-limit. Full YAML suite passes (273).

### DIFF
--- a/release-notes/VERSION
+++ b/release-notes/VERSION
@@ -16,6 +16,7 @@ implementations)
 
 3.3.0 (not yet released)
 
+#636: (yaml) Support `StreamReadConstraints.maxDocumentLength()` for YAML parser
 #699: (csv) Ability to modify columns in `CsvSchema.Builder` by name
  (requested by @OrangeDog)
  (fix by @seonwooj0810)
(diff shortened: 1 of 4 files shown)


cowtowncoder:
Re-created as #737

PR #730, opened 2026-09-12
Support StreamReadConstraints.maxDocumentLength() in Properties parser

pjfanning:
Completes #638 (which delivered `maxNameLength`; `maxDocumentLength` was left out because `java.util.Properties.load()` consumes the input). Same approach as TOML (#728) and YAML (#729): count at the input boundary.

Every `_createParser` variant in `JavaPropsFactory` funnels through `_loadProperties(Reader, IOContext)` (`InputStream` input via `Latin1Reader`, `byte[]`/`char[]` via `Latin1Reader`/`CharArrayReader`), so that single method now wraps the `Reader` in a small `ReadConstrainedReader` (`javaprop.io`) when `StreamReadConstraints.hasMaxDocumentLength()`. The wrapper accumulates characters read/skipped and calls `validateDocumentLength()` after each read; with no limit configured (the default) nothing is wrapped, so there's zero overhead. The `try-with-resources` close still closes the underlying reader through the wrapper.

Tests: `PropsDocumentLengthTest` — within/over the limit for `String`, `Reader`, `InputStream` and `byte[]` input plus direct streaming access; a limit below `Properties.load()`'s 8K read buffer; a single value longer than the limit (fails on document length, not string length); explicit no-limit. Full Properties suite passes (77).


cowtowncoder:
Re-created as #738.

PR #731, opened 2026-09-13
Apply TomlFactory's StreamReadConstraints to TreeTraversingParser it creates

pjfanning:
Follow-up to #728 for #430, using the `TreeTraversingParser` constructor added in FasterXML/jackson-databind#6216.

TOML parses into an `ObjectNode` and exposes it via databind's `TreeTraversingParser`, which used to take its `StreamReadConstraints` from the `ObjectReadContext`. So `TomlFactory`'s constraints only applied when the caller was an `ObjectMapper` (whose `DeserializationContext` delegates to the factory); with `tomlFactory.createParser(ObjectReadContext.empty(), …)` a configured `maxTokenCount` was silently ignored — unlike streaming parsers in the other modules, which always use constraints from their `IOContext`.

Fix is one line in `TomlFactory._createParser()`: pass `ctxt.streamReadConstraints()` explicitly. With this (plus #728), every `StreamReadConstraints` setting is enforced by the TOML backend regardless of how the parser is created.

Tests: `TomlTokenCountTest` — `streamReadConstraints()` returns the factory's instance; `maxTokenCount` enforced via `createParser(ObjectReadContext.empty(), …)` for `String`/`Reader`/`InputStream`/`byte[]`/`char[]`; still enforced via the mapper; within-limit document reads fully. 2 of 4 fail without the fix (the factory-path ones — the mapper path always worked, which is why this went unnoticed). Full TOML suite passes (858).

Requires a jackson-databind `3.3.0-SNAPSHOT` containing FasterXML/jackson-databind@02e30645 (#6216, merged).


pjfanning:
Folded into #728 (cherry-picked as 57d8aecf there) so the #430 work lands as one PR.

PR #738, opened 2026-09-18
Support `StreamReadConstraints.maxDocumentLength()` in Properties parser

cowtowncoder:
Re-created #730 (against `3.x`) to the `2.18` patch line, for the 2.18.11 release.

Completes #638 (which delivered `maxNameLength`; `maxDocumentLength` was left out because `java.util.Properties.load()` consumes the input). Same approach as YAML (#737, backport of #729): count at the input boundary.

Every `_createParser` variant in `JavaPropsFactory` funnels through `_loadProperties(Reader, IOContext)` (`InputStream` input via `Latin1Reader`, `byte[]`/`char[]` via `Latin1Reader`/`CharArrayReader`), so that single method now wraps the `Reader` in a small `ReadConstrainedReader` (`javaprop.io`) when `StreamReadConstraints.hasMaxDocumentLength()`. The wrapper accumulates characters read/skipped and calls `validateDocumentLength()` after each read; with no limit configured (the default) nothing is wrapped, so there's zero overhead. The `try-with-resources` close still closes the underlying reader through the wrapper.

Unlike the YAML backport, no exception unwrapping is needed here: `Properties.load()` lets the `IOException` propagate, and `_loadProperties` already declares `throws IOException`, so the `StreamConstraintsException` surfaces as-is.

### Note on the 3.x PR

#730's current head does **not** compile. The original commit added the wrap inside `_loadProperties(Reader r0, IOContext ctxt)`; the later merge of `3.x` (which brought in the #719 refactor splitting out `_readProperties(Reader r0, boolean closeReader)`) relocated those lines into `_readProperties`, where `ctxt` is not in scope. Worth fixing there. The `2.18` code still has the pre-refactor shape, so this backport applies the wrap in `_loadProperties(Reader, IOContext)` as originally written.

### Tests

`PropsDocumentLengthTest` (JUnit 4/`TestCase` style, matching the rest of the `2.18` suite) -- within/over the limit for `String`, `Reader`, `InputStream` and `byte[]` input plus direct streaming access; a limit below `Properties.load()`'s 8K read buffer; a single value longer than the limit (fails on document length, not string length); explicit no-limit. Full Properties suite passes (69).


release-notes/VERSION-2.x
```diff
@@ -20,6 +20,9 @@ Active Maintainers:
   `StreamReadConstraints.maxDocumentLength()` for TOML parser
  (requested by @pjfanning)
  (fix by @pjfanning)
+#638: (properties) Support `StreamReadConstraints.maxDocumentLength()` for
```
cowtowncoder: Wrong; need to change to #738.

